PR #6, opened 2016-11-18
Optimized Box and added all variants

lasalvavida:


pjcozzi:
Will wait for the discussion in #4.

Issue #16, opened 2016-12-26
Add glTF 1.1 models from the upcoming glTF tutorial


pjcozzi:
@javagl for the models in https://github.com/javagl/gltfTutorialModels, do you prefer:
* That we just link to that repo from here, or
* That you open a PR to move all those models into this repo and then perhaps add an appendix to the tutorial with a list of the models and direct links to each?

javagl:
I don't have a strong preference here. Conceptually, they are a bit different than the other models, because they are small, hand-crafted, feature-based examples, and some of them use a mixed structure (with embedded buffers but external shaders) for easier presentation in the tutorial. 

If they are moved to the official sample models repo, then they could
- either appear as "Test cases/models" in a dedicated list
- or be added to the list of existing models, possibly sorted by topic (e.g. the  "SimpleAnimation" before "Box Animated", the "SimpleSkin" before "Rigged Simple" etc.)

In any case, I agree to what you mentioned in the second point: It could be useful to link from the tutorial to the models. Either via a list in an appendix, or inlined: *"Here is an example ... it can be downloaded from the sample models repo at (link)"*. Similarly, one could add a link to the README of the models: *"SimpleSkin. Skinning is explained in the tutorial at (link)"*

If they should be added here, then just drop me a note, and I will create a PR to start filling the 1.1 directory :-)

pjcozzi:
> either appear as "Test cases/models" in a dedicated list

Good idea, let's do this.  Looking forward to the pull request!

javagl:
Done ("accidentally") as discussed in https://github.com/KhronosGroup/glTF-Sample-Models/pull/18#issuecomment-272644365

PR #18, opened 2017-01-04
Added simple test models for glTF 1.1

javagl:
See https://github.com/KhronosGroup/glTF-Sample-Models/issues/16


javagl:
These are the models from https://github.com/javagl/gltfTutorialModels

- All the models pass the validation, without errors. (Some generate warnings for unused mesh attributes, because the same meshes are used for multiple sample models, but this should be OK)
- The alignment of the accessor data is based on the size of the *full type*, and not only on the `componentType` (see https://github.com/KhronosGroup/glTF/issues/802 ). This is valid, regardless of whether the spec is relaxed in this point. *If* the spec is relaxed, new data may be generated, with fewer padding bytes
- Parts of the skinning spec for 1.1 are still being discussed - I added a note for this

pjcozzi:
Great, thanks @javagl.

* What is your plan with the gltfTutorialModels repo?  It would be good to just have these in one central place and perhaps remove that repo?
* I think the unused mesh attributes could be confusing to developers new to glTF.  Is it hard to clean up those models?  [gltf-pipeline](https://github.com/AnalyticalGraphicsInc/gltf-pipeline) could help but it would also make other changes you probably don't want.
* Relaxing the alignment is OK with me.  @lexaknyazev has also asked about this and I've been slow to respond.  I will comment in https://github.com/KhronosGroup/glTF/issues/802 now.

javagl:
Regarding the gltfTutorialModels:  
I'll probably keep it for a few days, maybe mark it as "outdated" with a link to the official repo, and delete it eventually

---

Regarding the unused attributes: 
I'm not sure. Right now, the *"SimpleMeshes"* shows an example with non-trivial meshes, with multiple attributes, but these attributes are not used until the *"AdvancedMaterial*" comes into play. This step was basically intended to break down the tutorial a bit: 
 - The [SimpleMeshes example section](https://github.com/javagl/glTF-Tutorials/blob/master/gltfTutorial/gltfTutorial_008_SimpleMeshes.md) shows the example (with unused attributes)
 - The [Meshes section](https://github.com/javagl/glTF-Tutorials/blob/master/gltfTutorial/gltfTutorial_009_Meshes.md) *explains* the example, and the meaning of the (unused) attributes
 - The [Advanced Material example section](https://github.com/javagl/glTF-Tutorials/blob/master/gltfTutorial/gltfTutorial_014_AdvancedMaterial.md) then finally **uses** these attributes

When omitting the unused attributes in the *"SimpleMeshes"* model, some minor restructuring may be necessary here. Also the *"SimpleMeshes"* model would then boil down to be "a trivial mesh attached to two nodes", and then the model itself may probably be omitted. 

As for the other models: I can remove the unused attributes, of course.

(BTW: I intended to extend the README of each model with a table or image showing the *exact* data layout - basically specific versions of the [image from the tutorial](https://github.com/javagl/glTF-Tutorials/blob/master/gltfTutorial/images/bufferBufferViewAccessor.png). The models are small enough to make this feasible, and it could helpful for understanding the layout and contents of the raw, binary data. I'd do this after the alignment issue is resolved). 

---

Regarding the alignment: 
This is not "critical", in the sense that *relaxing* the spec is strictly backward compatible. But, to epmhasize it once more: *If* it is relaxed, this can **never** be taken back. 
(The tutorial section about [accessors](https://github.com/javagl/glTF-Tutorials/blob/master/gltfTutorial/gltfTutorial_005_BuffersBufferViewsAccessors.md#accessors) until now does not take into account the alignment requirement at all, so this has to be updated in any case)

pjcozzi:
All sounds good.

> When omitting the unused attributes in the "SimpleMeshes" model

Perhaps leave it and just explain why it is there in the README.md.

Also, do you have access to this repo to make a 1.1 branch and then retarget this to the 1.1 branch so we can merge this into master at the same time the 1.1 spec is merged?

javagl:
I'll update the readme soon.

I don't have access to this repo, but can retarget the PR as soon as such a branch exists (then, possibly, already with the updates discussed here).

pjcozzi:
Here's the 1.1 branch: https://github.com/KhronosGroup/glTF-Sample-Models/branches/all

javagl:
@pjcozzi I have changed the target to 1.1 (and hope that GitHub is doing the right thing here under the hood...).

By the way, I created a small utility for generating the binary data of these sample models. I considered to extend this to add a functionality like "Show buffer layout..." in the jgltf-browser, but an not sure whether I'll integrate this. Regardless of that, it already offers the option to automatically create SVG/PNG images like this:
https://cloud.githubusercontent.com/assets/5597569/21744897/66777458-d51f-11e6-9270-098014e89e1a.png

I thought that it could be helpful to add these buffer layout images to the README of each sample model. Should I do this, or do you think that it may be confusing or too detailed? (For most of them, the image will be smaller than the one above, though...)

javagl:
And by the way: The last update fixes the warnings that referred to unused attributes (except for the "SimpleMeshes" example, where I added a statement in the README), all models pass the validation and can be displayed with JglTF. 

(Most other viewers still seem to have a problem with the missing `scene` property. Should I add this property? Actually, it should be *optional*, but most viewers/loaders seem to assume that it is there...)

pjcozzi:
> I thought that it could be helpful to add these buffer layout images to the README of each sample model. Should I do this, or do you think that it may be confusing or too detailed? (For most of them, the image will be smaller than the one above, though...)

Sounds cool to me.

pjcozzi:
>Most other viewers still seem to have a problem with the missing scene property. Should I add this property? Actually, it should be optional, but most viewers/loaders seem to assume that it is there...)

I think the viewers need to be fixed (yes, Cesium too!).

Until then, it could be fine to include the default scene so this is not stalled.  Up to you.  Or start opening pull requests into the engines. 😄

javagl:
Sorry, it seems like at some point in time, I got write permission to the Khronos repo. I did *not* have this from the beginning, and have *not* been notified about this. Have these permissions assigned accidentally, and should they be revoked? The last commits went to the main repo, and not into my fork. (May this somehow have been caused by switching the target?) Should I roll back the last three commits? Sorry if this causes inconveniences, I wasn't aware that I was able to push to origin at all...

pjcozzi:
@javagl we requested commit access for you at some point, not sure about the delay.  Given how quickly everything is coming together for glTF 1.1 and your tutorial and I would not worry about your commits landing in the main repo.

javagl:
OK, thanks, and sorry again: I'm usually pushing with git-gui to quickly review all changes, and it selects "origin" as the default push target - I'll be extra careful here in the future. 

So this pull request has become obsolete, and I'll close it.

BTW, regarding the `scene` property: According to the `gltf-test`, the loaders are now beginning to accept assets without a `scene`, so this may in the end turn out to be a good thing. But one point that may have caused some confusion here is the question about what happens when there are multiple `scenes`, but no default `scene`:

https://github.com/KhronosGroup/glTF/tree/master/specification/1.0#scenes :

> An additional property, `scene` (note singular), identifies which of the scenes in the dictionary is to be displayed at load time.

Conceptually, this would then be similar to the `cameras`. There is no property for a default `camera` (although one might consider adding one). Analogously, the currently active scene has to be maintained by the client application, and when there is no `scene`, then the application may choose any of the scenes (likely, but not necessarily, the first one...?)

I'll probably set up further test cases for handling *multiple* scenes.

pjcozzi:
I don't think the spec is clear here, but I am tempted to think that it would be to render all scenes; what other sensible default exists.  Please start a discussion in the main glTF repo.

javagl:
Opened https://github.com/KhronosGroup/glTF/issues/815 for further discussion.

javagl:
@pjcozzi Unless told otherwise, I would **delete** the current `1.1` directory later today, to avoid confusion for people who might think that there is an 1.1 version. 

Should I do this 
- directly here (as now, I seem to have the access rights)
- as a pull request?

The 2.0 version of the models could then be added when the dust has settled. 

(BTW: I'm working on basic PBR tests, but bringing this into a sound state will still involve some efforts....)

pjcozzi:
@javagl yes, please delete/rename `1.1` to `2.0` using a **pull request** so that everyone is aware.  In general, I rarely commit directly to master so that changes don't go unnoticed.

Great news on the PBR

Issue #3, opened 2016-11-16
BoxSemantics is the same as BoxTextured


javagl:
The [BoxSemantics](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/1.0/BoxSemantics) model is supposed to "include extra uniforms to test all uniform semantics". But it seems to be the same as [BoxTextured](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/1.0/BoxTextured): The `.gltf` file is the same (except for the different buffer file name), and the shaders are also the same in both cases.

(Probably, something was lost there when a manual editing was overwritten during an automatic conversion. I tried to dig into the history, but did not spot the culprit yet...)

pjcozzi:
Good eye, @javagl.

It most likely needs these: https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Specs/Data/Models/Box-Textured-Custom/CesiumTexturedBoxTest.gltf#L450-L476

@lasalvavida do you think this was one of your updates?

lasalvavida:
Probably. If I had to guess, I probably regenerated it from the COLLADA and didn't add the semantics back.

pjcozzi:
@lasalvavida fixed this in #23.

Thanks again @javagl, good eye!

PR #23, opened 2017-01-17
Fix box semantics

lasalvavida:
Fix for #3. I don't think there's anything to be done in glTF-MaterialsCommon for this since the technique no longer exists.


lasalvavida:
@pjcozzi

pjcozzi:
Thanks!

PR #27, opened 2017-01-25
Removed 1.1 sample models

javagl:
They will be added with version 2.0 after the
necessary updates.


pjcozzi:
I'm not the git expert, but would it be better to rename these to 2.0 and then squash the history so we don't bloat the repo?

We could then just leave this PR open until 2.0 is ready so these examples are discover-able, but there isn't a lot of churn in the repo.

javagl:
I thought that the main point was to avoid people seeing "1.1" here. The overall size of the test models is currently 1 MB, but of course, unnecessarily increasing the size and history should be avoided. I'm also not a Git expert (beyond the basic workflow), and am open for suggestions.

pjcozzi:
Sorry for the slow response, I think it is fine to have "2.0" models here - even in master I suppose - that are really "2.0 in progress" just so devs have something to work with as 2.0 finishes up.

So perhaps just `git mv` to rename 1.1 to 2.0.

javagl:
I locally started a branch "2.0" for the https://github.com/javagl/gltfTutorialModels where I intended to keep the 2.0-beta-state, and to create a PR where when they are "done". But I can also add them here earlier. So I'll try to
- update the models based on the current 2.0 (beta) spec state during the weekend
- create a PR to rename the directory here to 2.0
- create a PR with the models in the 2.0-beta-state here

pjcozzi:
Sounds perfect, thanks @javagl!

javagl:
It is very difficult to create a "2.0 state" of the models. Most of the models are "updated" now, but it's hard to say more about them than that they are "valid JSON". I tried to get a grip on PBR, but this will still require a *significant* amount of work. Assuming that every model that involved materials will have to use the `KHR_technique_webgl` extension, I can try to add them here ASAP, but am not sure whether this might not raise more questions than it will answer.

pjcozzi:
@javagl I think simple PBR materials that you could use will emerge quickly.

lexaknyazev:
@javagl Also, please don't rely on current state of `KHR_technique_webgl`.
Currently, main focus is on the core 2.0 spec to meet schedule.

PR #30, opened 2017-02-17
Renamed 1.1 to 2.0

javagl:
As a first step (instead of https://github.com/KhronosGroup/glTF-Sample-Models/pull/27 ), renaming the folder, to reduce confusion about the version number.

**The actual samples are not yet updated!**

*Some* of them have been updated (currently only in https://github.com/javagl/gltfTutorialModels/tree/2.0 ), but the states of the `technique` and `material` spec (and things like skinning) are not yet stable enough. I also started creating further "tutorial models" for PBR (roughly like https://github.com/javagl/gltfTestModels/tree/master/SimpleSpheres), but these are also still in a *very* early state. The shader implementations require some more efforts, and I'm not sure how much time I can invest for this in the near future.


pjcozzi:
Thanks @javagl!

PR #32, opened 2017-02-21
Cleaned up PBR samples

bghgary:
- Formatted JSON
- Output one bin file instead of a bunch of bin files
- Named all the images


pjcozzi:
Thanks @bghgary!

Issue #34, opened 2017-02-24
bufferView#target of BoomBox sample


kyasbal-1994:
Is this valid data?

https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/2.0/BoomBox/glTF/BoomBox.gltf#L93

I think this bufferView.target should be WebGLRenderingContext.ELEMENT_ARRAY_BUFFER not ARRAY_BUFFER.

bghgary:
Thanks for the report. I will fix this.

pjcozzi:
@kyasbal-1994 thanks again for the report.  @bghgary fixed this in #35.

PR #35, opened 2017-02-26
GLB v2, bufferView target fix, and updated to conform to latest 2.0 s…

bghgary:
…chema

Normal map was updated to fix some compression artifacts


bghgary:
Addresses #34

pjcozzi:
Thanks @bghgary!

PR #40, opened 2017-04-02
 Norbert's glTF 2.0 sample models

McNopper:


pjcozzi:
Thanks @McNopper.

@bghgary do you want to double check these before we merge?

bghgary:
@pjcozzi I looked through the JSON files. The only thing that jumped out were the `TANGENT` and `BITANGENT` attributes on mesh. Do we want our sample models to include them?

emackey:
Is this going to add conflicts with #28?

McNopper:
Maybe I am not up to date anymore, but I thought either TANGENT4 or TANGENT/BITANGENT can be used:
https://github.com/KhronosGroup/glTF/issues/812

bghgary:
The WG had some concerns with stored tangents and morph targets exceeding the mesh attribute limits of WebGL. The current plan is to not add stored tangents support to the core spec.

I'm currently trying to define what is the right math for calculating the tangent frame at runtime.

@McNopper Is it easy for you to remove the tangents/bitangents for now?

McNopper:
My loader can currently do both - TANGENT4 and TANGENT/BITANGENT.
When exporting, I currently store TANGENT/BITANGENT but I can default to TANGENT4.
So, yes, these are just a few lines from my side.
Let me fix this and I will do another pull request.

McNopper:
Okay, made another pull request:
https://github.com/KhronosGroup/glTF-Sample-Models/pull/42
Using TANGENT4 as default now.
Think I got the math right (using http://crytek.com/download/izfrey_siggraph2011.pdf).

PR #41, opened 2017-04-04
Migrate COLLADA folders out of 1.0 folder

lasalvavida:
@pjcozzi 

This is a piece of #28. It just moves the COLLADA folders out of 1.0 into the root and adds a README explaining that some of the glTF models are generated from these source COLLADA models.


pjcozzi:
@lasalvavida just one small request: instead of naming this directory `collada` can you name it something more generic like `originalModels` or `sourceModels` since (1) we don't want users mistaking this to mean some kind of COLLADA/glTF relationship, and (2) we may have models in other formats like FBX.

lasalvavida:
Agreed, I've changed the folder to `sourceModels`

lasalvavida:
Tweaked the README slightly to reflect this as well

pjcozzi:
Thanks @lasalvavida!

PR #42, opened 2017-04-05
Norbert's glTF 2.0 samples - using TANGENT4.

McNopper:


bghgary:
> Using TANGENT4 as default now.
Think I got the math right (using http://crytek.com/download/izfrey_siggraph2011.pdf).

Sorry about this, I think we are not on the same page.

A couple of things:
1. TANGENT4 is not QTangent. It's tangent (XYZ) + handedness (-1 or 1).
2. What I meant was remove the tangents completely, but there is an issue here. We haven't defined the tangent math when they are not provided, so there is no way to generate a correct normal map right now.

Note: The models I submitted (BoomBox, Corset, and Lantern) do have normal maps but they don't have tangents. The normal maps use mikktspace (as they came from Unity), but as far as I'm aware, there isn't an easy way to calculate mikktspace at runtime. Thus, the normal map for these models will not render correctly in BabylonJS or ThreeJS as they both use a derivative-based tangent calculation in the shader. I'm working to figure out what a good standard way to calculate tangents at runtime. Once that's defined, then we can submit proper normal maps that work everywhere.

We have a couple of options:
1. Wait for the tangent math to be defined, don't specify tangents at all, and update the normal maps to match the math.
2. Commit these with TANGENT4 (tangent + handedness) but realize they will need to become option 1 later.
3. Submit just the models without normal maps for now and submit the models with normal maps later.

McNopper:
Ah okay, did not know that. I had this discussion with these QTangents in mind,
I will go for option 2.
WIll do my changes,

bghgary:
Looks good to me now.

McNopper:
Oops, you were fast, Had no time to inform you :-)

McNopper:
@pjcozzi Okay for you as well?

pjcozzi:
Thanks @McNopper @bghgary!

PR #45, opened 2017-04-11
Updated simple test models to glTF 2.0

javagl:
Most of the test models have been updated to 2.0. They may have to be reviewed, but in general, they should be compliant to the spec. A validation might reveal details that have to be fixed. 

Only the AdvancedMaterial, SimpleOpacity and SimpleSkin model updates are still pending. These updates will depend on the MaterialCommon extension and the latest changes of skinning, which are not integrated yet.


pjcozzi:
@lexaknyazev about when do you think the validator will be ready for us to test these?

javagl:
No haste - Alexey is probably busy with more urgent tasks right now. I'll also review the models again ASAP, which goes hand in hand with checking the spec changes. The "big" changes actually affect the models that are not updated yet (materials and skinning), and I'll try to create further test models (morphing, sparse accessors...), so there's still a lot of work to do to consolidate all this anyhow.

pjcozzi:
Fixes most of #53, see https://github.com/KhronosGroup/glTF-Sample-Models/issues/53#issuecomment-298789600.

cx20:
@javagl Is the position attribute of SimpleMaterial.gltf in this PR correct?
It doesn't seem to have a parameter corresponding to position attribute.

https://github.com/javagl/glTF-Sample-Models/blob/master/2.0/SimpleMaterial/glTF/SimpleMaterial.gltf#L86-L110

/cc @takahirox

Relate : https://github.com/mrdoob/three.js/issues/11326#issuecomment-301669751

javagl:
@cx20 That's hard to say. It may not be correct. I think it once *was* correct. At least, once it was *supposed* to be correct (in the future) : https://github.com/KhronosGroup/glTF/issues/789#issuecomment-265509673 

The `KHR_technique_webgl` extension is not finalized, and the `SimpleMaterial` model explicitly declared to be not updated yet. But I'll simply remove this model, having it in there does not make sense.

lexaknyazev:
@javagl 
Couple fixes before merging:
- must be `node` instead of `id`:
  - [glTF/AnimatedTriangle.gltf#L38](https://github.com/javagl/glTF-Sample-Models/blob/master/2.0/AnimatedTriangle/glTF/AnimatedTriangle.gltf#L38)
  - [glTF-Embedded/AnimatedTriangle.gltf#L38](https://github.com/javagl/glTF-Sample-Models/blob/master/2.0/AnimatedTriangle/glTF-Embedded/AnimatedTriangle.gltf#L38)
- `bufferView.byteStride` must be defined when two or more accessors use the same buffer view:
  - [glTF/SimpleMeshes.gltf](https://github.com/javagl/glTF-Sample-Models/blob/master/2.0/SimpleMeshes/glTF/SimpleMeshes.gltf)
  - [glTF-Embedded/SimpleMeshes.gltf](https://github.com/javagl/glTF-Sample-Models/blob/master/2.0/SimpleMeshes/glTF-Embedded/SimpleMeshes.gltf)

PR #50, opened 2017-04-18
Metal-Rough PBR test models

### LINKED ISSUE #47: Simple sample model for testing PBR
If there was simple model just for testing pbr like below, it would be great. 

And it is very good for comparing 2 libraries supporting PBR.

emackey:
Added two sample models, one to test metalness vs. roughness, and another to test what happens when normal map UV coordinates are rotated into different orientations.

Fixes #47.


pjcozzi:
Great, thanks!

PR #58, opened 2017-05-09
Add two animated morph target samples

HowardWolosky:
Adds two new sample files that demonstrate morph targets and the
animation of those targets.

The first sample is a simple cube with two morph targets:
  * `thin` moves four vertices to create a thin cube
  * `angle` moves two vertices to create a ramp

The second sample is a simple sphere with two morph targets that are
more complicated:
  * `blob` moves many vertices to create an amorphous blob
  * `ship` moves even more vertices to create something that resembles a spaceship.

In both cases, there is an animation that animates between the two morph targets attached to the object.

The [Babylon.js sandbox](https://sandbox.babylonjs.com/) is a great way to try these models out.

Thanks to @BeardedGnome and @deltakosh for the [Babylon.js](https://github.com/BabylonJS/Babylon.js) changes needed to both support morph targets in general, as well as to the glTF-specific loader.

Thanks also to @bghgary for the [Unity Exporter](https://github.com/bghgary/glTF-Tools-for-Unity) help.


HowardWolosky:
Whoops.  Fixed

bghgary:
@pjcozzi @emilian0 @lexaknyazev Does this look okay to you?

pjcozzi:
I have not worked on morph targets in any detail so up to @emilian0 @lexaknyazev.

I would suggest that we merge this soon to spur additional interest in implement morph targets and iron out and data/spec issues.

lexaknyazev:
While the spec is explicit on attribute accessor formats for the base mesh, it doesn't define them for morph targets. This seems to be trivial for `POSITION` and `NORMAL` attributes: just use the same as the base mesh uses.

But it's not that simple for `TANGENT`: the fourth component means handedness and there's no sense in "interpolating" handedness between morph targets. Thus, morph target's `TANGENT` accessor should be of  `VEC3` type.

So, PR's models lgtm, but the spec needs to be updated.

@bghgary Could you please add notes about `TANGENT` morph target attribute format and usage?

bghgary:
Yes, we noticed the same thing while implementing. I will make this update to the readme.

Issue #59, opened 2017-05-12
Validation errors in collada2gltf 2.0 models


lexaknyazev:
As of ef49777e5e63632b089b238bffe4515dbe8e1325.

- `camera.type` is absent;
- `byteStride` changes from 2.0 aren't implemented;
- integers in `accessor.min/max` are written with `.0`;
- length in GLB header is 8 bytes fewer than it must be;
- `KHR_binary_glTF` mustn't be written;
- IBM matrices mustn't be located in the same bufferView as vertex attributes;
- `skin` schema isn't updated.

@lasalvavida @pjcozzi

PR #61, opened 2017-05-12
Updates from validation

lasalvavida:
@lexaknyazev fixes for #59


lexaknyazev:
@lasalvavida 
Great! Most issues are resolved. Remaining ones:
- `bufferView.byteStride` must be defined when two or more accessors use the same buffer view.
- `skin.joints` must be an array of nodes indices, `node.jointName` mustn't exist.

lasalvavida:
> bufferView.byteStride must be defined when two or more accessors use the same buffer view.

One question here, the spec says:

> Buffer view could have byteStride property ... This field is defined only for buffer views that contain vertex attributes.

So, is this is only true for vertex attribute bufferViews correct? Is the intent here to have vertex attributes of different sizes split across multiple bufferViews such that they all have the same byteStride, or is it still acceptable to pack vertex attributes with different byte strides into a single bufferView with a declared byteStride of 0?

lexaknyazev:
`byteStride` can't be `0`, so all accessors in a bufferView explicitly have the same byteStride.

lasalvavida:
Okay, I'll update my implementation to match, just so you know the spec still says:

> bufferView.byteStride
> The stride, in bytes, between vertex attributes or other interleavable data. When this is zero, data is tightly packed.

lasalvavida:
And the default value for byteStride is still listed as zero

lexaknyazev:
Please, refer to JSON files for the most current properties definitions:
https://github.com/KhronosGroup/glTF/blob/2.0/specification/2.0/schema/bufferView.schema.json#L23-L31

The reference part of README.md is outdated.

lasalvavida:
Updated

lexaknyazev:
Couple more small tweaks:
- GLB versions mustn't contain `image.width/height`. There're no such fields in glTF 2.0.
- Default `node.matrix` value (identity matrix) should be omitted. E.g., see how it clutters **ReciprocatingSaw**.
- `camera.perspective.aspectRatio` can't be zero (see **VC** model).

lasalvavida:
> camera.perspective.aspectRatio can't be zero (see VC model).

This actually comes from the COLLADA which appears to be incorrect. `aspect_ratio` is a required property if only a single `xfov` or `yfov` is present.

That being said, it looks like I can just omit this and it defaults to the canvas aspect ratio, correct?

lexaknyazev:
> That being said, it looks like I can just omit this and it defaults to the canvas aspect ratio, correct?

Yes.

lasalvavida:
Updated

lexaknyazev:
@lasalvavida 
Were models updated after https://github.com/KhronosGroup/glTF-Sample-Models/issues/63#issuecomment-301344107 change? JSON's min/max values seem to be the same.

lasalvavida:
Hmm... COLLADA2GLTF is using floats internally now where it was using doubles before; rapidjson doesn't have a write for float, only double, so maybe it has something to do with that. I'll look into it

pjcozzi:
@lasalvavida - small request: please squash commits before this is merged to keep the repo size down, thanks!

lexaknyazev:
Looks like ef89bd4032d5962920f9743eb5b1f6a9f5d33d1c accidentally removed many files from the tree.

@lasalvavida 
Could you please bring them back or just open a new PR with collada2gltf-only updates?

lasalvavida:
Closed, continued in #67

PR #66, opened 2017-05-19
Add tools for generating sample models

lasalvavida:
@pjcozzi, thought this might be useful to add; it's how I generate variants


pjcozzi:
Can you provide a bit more context in the README.md?  For example a link to https://github.com/KhronosGroup/glTF-Sample-Models, the options, and, generally, why the script exists compared to just using COLLADA2GLTF directly.

lasalvavida:
I've fleshed out the README a bit. 

> For example a link to https://github.com/KhronosGroup/glTF-Sample-Models

What did you mean by this?

pjcozzi:
>> For example a link to https://github.com/KhronosGroup/glTF-Sample-Models

> What did you mean by this?

Simply a link to the glTF Sample Models.  COLLADA2GLTF users may not know the samples repo exists so they may not understand what the script is for.

Just some context.

emackey:
I don't want to step on any toes here, because I know this PR represents a ton of work and fine-tuning.  But I have concerns as to how much this will still be needed after 2.0 becomes finalized.  Ideally the spec should settle down, and constant automated re-conversion will become a thing of the past, no?

Also, how well supported are metal-rough source models here?  If I have a source model in COLLADA that uses an occlusion-roughness-metallic texture, what will this script do?  Will it attempt to create a spec/gloss version, and will that version actually contain correctly converted spec/gloss maps?

tools/README.md
```diff
@@ -0,0 +1,25 @@
+# glTF Sample Model Tools
+## generateSampleModels.py
+Python script to generate glTF models from COLLADA [source models](../sourceModels) using [COLLADA2GLTF](https://github.com/KhronosGroup/COLLADA2GLTF).
+
+Place a built COLLADA2GLTF binary in the tools folder and execute the script.
```
emackey: This clutters up the repo with files that shouldn't be submitted.  Perhaps the user could set an environment variable or something to indicate where they keep their copy of COLLADA2GLTF, so that when it gets updated, this script always gets the update.

pjcozzi:
> This PR is in the sample models repo. How could they be here if they didn't know it existed?

hahaha, you are right.  I am spread across too many repos.  Ignore me.  😆

pjcozzi:
@emackey what do you suggest here?  Pass on this as we won't need it once everything is updated to 2.0?

emackey:
Yeah, at most I would park this on a side branch for future reference.  But I don't think it should ship on the master branch in this form.

pjcozzi:
OK, could you please park and close this?

emackey:
OK, this is parked on a branch called `tools` in the main repo.

PR #67, opened 2017-05-19
Squashed updated from validation

lasalvavida:
@lexaknyazev continued from #61 for #59.


lasalvavida:
Also, @lexaknyazev, is there any chance we could get a published dev branch for the 2.0 validator? It would help me a lot in testing gltf-pipeline.

Issue #74, opened 2017-05-24
Update SpecGloss models from COLLADA2GLTF


lexaknyazev:
- Schema syntax bug: https://github.com/KhronosGroup/COLLADA2GLTF/issues/43.

- Models without textures (`2CylinderEngine`, `Box`, `BoxAnimated`, `BrainStem`, `Buggy`, `GearboxAssy`, `ReciprocatingSaw`, `RiggedFigure`, `RiggedSimple`, `WalkingLady`) don't have SpecGloss material at all despite having `glTF-pbrSpecularGlossiness` directory.

- All SpecGloss models from COLLADA2GLTF use base64-encoded buffers and images.

emackey:
We're deprecating the old SpecGloss extension in favor of `KHR_materials_specular`.

As far as I know, all of the old SpecGloss samples were merely converted from metal/rough sources, or old-fashioned non-PBR diffuse/specular sources.  None of them were taking advantage of the PBR-enabled spec/gloss workflow beyond what metal/rough can do.  There is a newer Specular Test model that tests `KHR_materials_specular`.

Perhaps all of the `glTF-pbrSpecularGlossiness` folders should be removed.

PR #76, opened 2017-05-26
Regenerated sample models from COLLADA2GLTF updates

lasalvavida:
These were generated with: https://github.com/KhronosGroup/COLLADA2GLTF/pull/50

so there should be no further churn in the binaries unless something actually changes in the data.


lasalvavida:
I see that there is a unused node still in MilkTruck. Looking into it.

bghgary:
```JSON
     "extensionsRequired": [
         "KHR_materials_pbrSpecularGlossiness"
     ],
```

Is there a specific reason why this is in the pbrSpecularGlossiness models? The metallic-roughness materials are present. Implementations should be able to fallback to metallic-roughness if they don't have specular-glossiness support.

lasalvavida:
> Is there a specific reason why this is in the pbrSpecularGlossiness models? The metallic-roughness materials are present. Implementations should be able to fallback to metallic-roughness if they don't have specular-glossiness support.

That sounds reasonable to me

bghgary:
Looks like some of the models don't have specular-glossiness materials even though it's in the spec-gloss folder and has extensionsUsed. For example, [2.0/Box/glTF-pbrSpecularGlossiness/Box.gltf](https://github.com/lasalvavida/glTF-Sample-Models/blob/c2238ec684c95eac1aa8cee33944be63a2f45dda/2.0/Box/glTF-pbrSpecularGlossiness/Box.gltf)

bghgary:
Also, at least the pbrSpecularGlossiness models seem to have base64 encoded data even though there is also side-car files.

For example:

[2.0/Box/glTF-pbrSpecularGlossiness/Box.gltf](https://github.com/lasalvavida/glTF-Sample-Models/blob/c2238ec684c95eac1aa8cee33944be63a2f45dda/2.0/Box/glTF-pbrSpecularGlossiness/Box.gltf)
> ```JSON
> "buffers": [
>         {
>             "byteLength": 648,
>             "uri": "data:application/octet-stream;base64,..."
>         }
>     ],
> ```

Yet, there is a Box0.bin next to the file.

lasalvavida:
> Also, at least the pbrSpecularGlossiness models seem to have base64 encoded data even though there is also side-car files.

Good catch, that looks like an error from #66. Will fix

lexaknyazev:
@lasalvavida Will this PR cover all issues from #74?

lasalvavida:
Updated. And yes, this should address everything from #74.

lexaknyazev:
@lasalvavida Looks like you've found validator's bug. It crashes when GLB file refers to non-existent external URI.

lasalvavida:
Looks like I was a little overzealous with my separate flags

bghgary:
Ah, sorry @lexaknyazev, didn't realize you already mentioned these issues in #74.

lasalvavida:
Updated

2.0/Box/glTF-pbrSpecularGlossiness/Box.gltf
```diff
@@ -112,8 +112,26 @@
...
+            "extensions": {
+                "KHR_materials_pbrSpecularGlossiness": {
+                    "diffuseFactor": [
+                        0.800000011920929,
+                        0.0,
+                        0.0,
+                        1.0
+                    ],
+                    "specularFactor": [
+                        0.20000000298023225,
```
bghgary: Where did these numbers come from? I'm expecting 0.04 for specularFactor since metallicFactor is 0.

lexaknyazev: Looks like classical (non-PBR) COLLADA's `<specular>` is being used here:
https://github.com/lasalvavida/glTF-Sample-Models/blob/2.0-fixes/sourceModels/Box/Box.dae#L27

lexaknyazev: @bghgary 
Btw, should such case (both models defined for the same material, `specularFactor` isn't 0.04) be a validation warning?

bghgary: I see. This is the conversion (or lack of) from SG to MR. Do we want SG to look different from MR in these models?

lexaknyazev: > Do we want SG to look different from MR in these models?

- Since these are sample models, we can expect that they will be used as "proper" examples. So results must be predictable.
- If the goal of using both materials in one asset is to achieve better compatibility/performance, they should look "similar".
- We should provide additional samples to highlight differences between MR and SG.

bghgary: > a validation warning

Perhaps. It will be tough for the validator to verify this until you also include math for converting SG to MR (which may or may not be the same for all implementations). Also, would you verify only factors and not textures? I think it will be difficult for the validator to validate visual issues.

lexaknyazev: > I think it will be difficult for the validator to validate visual issues.

Sure. Implementing that (esp. software texture sampling) isn't on the roadmap at the moment. But such asset should be invalid, if we expect spec-defined relations between SG and MR.

bghgary: >  But such asset should be invalid, if we expect spec-defined relations between SG and MR.

+1

lasalvavida:
Is there something specific you'd like me to do here, or is this okay to merge?

bghgary:
I think merging this is an improvement, so I don't think we should stop merging because of the SG vs. MR issue. Maybe we can file a separate issue on that?

lexaknyazev:
We need to regen models one more time to fix binary padding in some of them.
Also, @lasalvavida could you move WebGL-technique fields to an extension object? 
Additional properties are allowed by schema only for forward compatibility, not for extensions.

PR #84, opened 2017-06-21
Box example with interleaved position and normal attributes

AngeloReppucci:
While testing models with interleaved attributes i found that different engines make errors. In particular, i tested babylon.js and three.js and either are unable to display correctly this kind of model. 
The difference compared to the original Box example are in the layout of the buffer.

In the original sample, the layout is the following (where P=Position, N=Normal, I=Index) with a bufferView.stride = 12 for attributes (also if to my understanding it must be 0 in this case) :
 
PPPPPPPPPPP....................NNNNNNNNNNN................IIIIIIIIIIIIIIIIIIIIIIIIIIIII

The model of this PR has the following buffer layout with a bufferView.stride = 24 for attributes:

PNPNPNPNPNPNPNPN........................IIIIIIIIIIIIIIIIIIIIIIIIIIIIIII


lexaknyazev:
> also if to my understanding it must be 0 in this case

`byteStride` of `0` isn't allowed. For vertex data, it could be omitted only when `bufferView` is being used by only one accessor.

Here's relevant spec part:
https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/schema/bufferView.schema.json#L23-L31

---

How was GLB version produced? It has wrong padding in JSON chunk.

AngeloReppucci:
@lexaknyazev, thanks for the clarification byteStride.

I manual edited the glb and forgot to update the length in the header (now is 1631 vs 1632 of the original version).

Attached there is the corrected version. 

[Box_Interleaved.zip](https://github.com/KhronosGroup/glTF-Sample-Models/files/1093053/Box_Interleaved.zip)

lexaknyazev:
@AngeloReppucci 
Length of each chunk must be a multiple of `4`, so total length should be 1632, and JSON chunk's length should be 956 (including 3 padding bytes).

For more detailed explanation see
https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#chunks

I've updated file in this PR.

Thanks for contribution!

AngeloReppucci:
@lexaknyazev, thanks for the explanation.

Happy, to give my contribution to the community.

alteous:
Is there any reason why the corresponding .gltf is called Box_Interleaved.gltf rather than BoxInterleaved.gltf? All other example models follow this pattern.

lexaknyazev:
@alteous 
Good catch! Fixed in #88.

PR #91, opened 2017-06-28
Add texture coordinate test.

### LINKED ISSUE #82: Add a texture coordinates test
Similar to the [TextureSettingsTest](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/TextureSettingsTest), there should be a "texture coordinates test".

Of course, in some way, this will be covered by all other models that involve textures, but I think that it might be nice to have a **dedicated**, *minimalistic* test case for this distressingly common issue.

In order to avoid any further confusion:Can anybody confirm that 

- putting **this** texture on a unit square

    test

- using **the texture coordinates matching the vertex positions, as shown in the image itself (!)**
- rendering the quad with with the default camera 

will result in the quad appearing exactly like the texture?

(Particularly: Is it right that the texture will **not** appear to be *flipped* ?)

emackey:
Fixes #82.

See the included model's new [README.md](https://github.com/emackey/glTF-Sample-Models/tree/56a9556d0a6460e9a05c60106186f605805145a6/2.0/TextureCoordinateTest).


pjcozzi:
Very nice, @emackey!

pjcozzi:
Could you also add this to the main README.md?

PR #92, opened 2017-06-28
Add Damaged Helmet model.

emackey:
/cc KhronosGroup/glTF-WebGL-PBR#35

It was requested that this model also be added here to the sample model repo.


emackey:
No textures were vertically flipped during the conversion of this model.  But the original was old enough that metalness was still in the red channel, so that was manually moved to the blue channel.

pjcozzi:
Thanks @emackey!

pjcozzi:
Could you also add this to the main README.md?

PR #98, opened 2017-07-13
Smooth out the seam in Damaged Helmet.

emackey:
Fixes KhronosGroup/glTF-WebGL-PBR#38.  @abwood was pestering me this morning about the visible seam in the helmet.

The original model had come in early draft glTF format, without the seam but not aligned to what later became the final release of glTF 2.0.  I corrected the schema problems by importing and then re-exporting the model from Blender, and somehow in that process a seam was introduced between the two halves of the model (the textures cover only half the model, the other side is a mirror image).

In this PR, I've re-loaded the Blender scene, run "Remove Doubles" on all vertices to remove all seams, run a 28-degree Edge Split modifier to regenerate intentional seams, and re-exported to glTF.  The updated blend file is included here along with updated glTF files.


abwood:
The seam appears to be fixed with these changes.  :+1: 
helmet_fix

emackey:
Thanks @abwood.

@pjcozzi can you merge?

pjcozzi:
@emackey do you want commit access to this repo?

Issue #77, opened 2017-06-01
Consider adding some kind of index file


emackey:
I'm thinking about using this repo as a git submodule in other projects that could use sample models, specifically glTF-WebGL-PBR, and possibly @cx20's gltf-test if he's willing.

To make the models more discoverable, in an automated fashion, it might be nice to have a JSON file included with this repo that lists each of the available samples, along with its name, description, screenshot URL, and available variants (embedded, binary, pbrSpecGloss, etc).

I'm imagining a NodeJS script or similar that would scan the 1.0 and 2.0 folders and compile a list of this information, and save it into some generic JSON structure.  The gltf-test project already has [the beginnings of such a structure](https://github.com/cx20/gltf-test/blob/master/sampleModels/model-index.js#L4-L24) hard-coded into it, but I'm picturing one that's a little more fleshed out.  The file would likely be called `model-index.json` in the root folder and could be rebuilt from the command line whenever the sample models change.  Such a file could be used or ignored at will by any software looking to load sample models.

Thoughts?

pjcozzi:
As long as we keep the barrier-to-entry low for submitting sample models, e.g., generate the .md or an HTML file from this and the contributor only needs to provide the JSON, I think it could work.

Perhaps one .json per file and then a bot running in CI to generate the master .json.

bwasty:
I've written a Python script to generate two index files (for 1.0 and 2.0) from the existing data: https://github.com/KhronosGroup/glTF-Sample-Models/compare/master...bwasty:generate_index

The reason actually is to get rid of a git submodule, because it [doesn't really work that well](https://github.com/alteous/gltf/issues/38). The idea is to download individual models via `raw.githubusercontent.com`.

Not sure about the .json per file and CI suggestions - perhaps this could be enough as a first step?

emackey:
Thanks @bwasty, I'll take a closer look soon.  One thought off the top of my head, the json files look nice and tidy now with some baked-in assumptions about the folder structure:

```
 {
   "name": "BarramundiFish", 
   "screenshot": "screenshot/screenshot.png", 
   "variants": [
     "glTF", 
     "glTF-Binary", 
     "glTF-Embedded"
   ]
 }, 
```

But this requires the user to be aware of some strange rules, for example the rule that the "glTF-Binary" variant is typically `.glb` instead of `.gltf`.  Do you think it would be worth expanding this out?

```
 {
   "name": "BarramundiFish", 
[...]

PR #112, opened 2017-08-11
Add model index files + generator

bwasty:
Adds `model-index.json` files and a Python script to generate them. See #77 for background.

### DIFF
--- a/generate-index.py
+++ b/generate-index.py
@@ -12,13 +12,11 @@ def generate_index(root):
         os.chdir(model)
         model_contents = os.listdir(".")
         gltf_variant_dirs = [d for d in model_contents if d.startswith("glTF")]
-        variants = []
+        variants = {}
         for variant_dir in gltf_variant_dirs:
-            if (not os.path.exists("{}/{}.gltf".format(variant_dir, model))
-                    and not os.path.exists("{}/{}.glb".format(variant_dir, model))):
-                print "WARNING: did not find model file for {} {}".format(variant_dir, model)
-                continue
-            variants.append(variant_dir)
+            model_file = [f for f in os.listdir(variant_dir)
+                          if f.endswith(".glb") or f.endswith(".gltf")][0]
+            variants[variant_dir] = model_file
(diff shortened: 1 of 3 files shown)


pjcozzi:
Really cool, thanks @bwasty!

@emackey did you want to look at this?

@lexaknyazev is this easy to connect to CI?

bwasty:
@emackey I've just rebased and re-generated the indexes since I want to use them for a new viewer I'm working on.

emackey:
Thanks @bwasty.

generate-index.py
```diff
@@ -0,0 +1,36 @@
...
+        os.chdir(model)
+        model_contents = os.listdir(".")
+        gltf_variant_dirs = [d for d in model_contents if d.startswith("glTF")]
+        variants = {}
+        for variant_dir in gltf_variant_dirs:
+            model_file = [f for f in os.listdir(variant_dir)
+                          if f.endswith(".glb") or f.endswith(".gltf")][0]
+            variants[variant_dir] = model_file
+        if "screenshot" not in model_contents:
+            print "WARNING: no screenshot found for {}".format("model")
```
emackey: So I just merged this, and then I found a problem.  On line 21 here, `model` should not be in quotes.  Also, Python 3.x requires the `print` arguments to be in parentheses, as it changed from a statement to a function.  Can you open a new PR for that?  Thanks!

generate-index.py
```diff
@@ -0,0 +1,36 @@
...
+        index.append({
+            "name": model,
+            "variants": variants,
+            "screenshot": "screenshot/" + os.listdir("screenshot")[0]
```
emackey: Also, without an `else` for the above conditional, the missing screenshot folder will fail the script here.

generate-index.py
bwasty: Ok, will do.

emackey: Addressed in #168.

generate-index.py
emackey: Addressed in #168.

PR #122, opened 2017-10-10
Remove broken WalkingLady and SmilingFace glTF 2.0 output.

### LINKED ISSUE #120: Move invalid models to a subdirectory or branch?
Should we move invalid glTF models such as WalkingLady (#83) and SmilingFace (?) to a separate subdirectory or branch?

Having them in the main samples area without them working in the popular engines is causing some confusion.

donmccurdy:
Fixes #120.

@lasalvavida @javagl is this OK? It looks like you are already getting close on a fix for the WalkingLady model.


pjcozzi:
Thanks @donmccurdy.

Merging, see https://github.com/KhronosGroup/glTF-Sample-Models/issues/120#issuecomment-335554211

PR #123, opened 2017-10-11
Fix models created by glTF Tools for Unity

bghgary:
Re-export models from glTF Tools for Unity
- Fix incorrect vertex tangent values
- Remove min/max properties for accessors that don't require them
- Remove incorrect color factor in Lantern model

This addresses https://github.com/KhronosGroup/UnityGLTF/issues/75#issuecomment-335616894


gildor2:
Thank you for this PR. I've tested updated models and now tangent vectors appears correctly.

lexaknyazev:
@bghgary 
Please, keep min/max values for morphed `POSITION` attribute.

PR #125, opened 2017-10-26
Update glTF-Tools-for-Unity exported models to +Z forward

bghgary:
See https://github.com/KhronosGroup/glTF/issues/1043#issuecomment-339132697


emackey:
Thanks @bghgary!

Issue #109, opened 2017-08-04
consider binary glTF using KHR-MC extension


andreasplesch:
I believe the binary glTF (.glb) files are generated from glTF using gltf-pipeline, both versions 1.0 and 2.0. It further appears that default options for gltf-pipeline were used. [It would be helpful to explicitly know if this is the case].

The default for gltf-pipeline for KHR-extension using input models, I believe, is to 'optimize' output to not use the extension and include shaders instead. So the generated binary glTF files would stop using the KHR (MaterialsCommon) extension.

At least, the provided v.1.0 .glb files do not appear to use the KHR-extension, from what I could determine. I suspect the v.2.0 .glb files also do not use the KHR-extension.

For testing purposes, please consider adding a glTF-binary-MaterialsCommon version for the models which have already a glTF-MaterialsCommon version. This potentially could be automated by travis.

If there is interest, I could take a stab at this, using current glTF-pipeline with the -kmc.enable option and assuming it works correctly.

donmccurdy:
Hi @andreasplesch, note that the `KHR_` prefix now applies to [multiple extensions](https://github.com/KhronosGroup/glTF/tree/master/extensions/Khronos), and that the KHR_materials_common extension is [likely to be replaced with KHR_cmnBlinnPhong](https://github.com/KhronosGroup/glTF/issues/947). 

For glTF 2.0, there is currently no support for custom shaders at all. It is coming later, in a separate extension, but will probably not be the default from gltf-pipeline. 

So, I think that when gltf-pipeline is fully updated for glTF 2.0, and the KHR_cmnBlinnPhong extension is finalized, this will no longer be a problem. Looking at the very end of discussion on https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/272, there will be an update to gltf-pipeline soon.

javagl:
>  For glTF 2.0, there is currently no support for custom shaders at all. It is coming later, in a separate extension, but will probably not be the default from gltf-pipeline.

Just a side note: This extension will likely be based on the glTF 1.0 ways of doing it, and will likely be available as https://github.com/KhronosGroup/glTF/tree/master/extensions/Khronos/KHR_technique_webgl , but as Don said, this is **currently only a stub**.

andreasplesch:
Thanks for the update. Since glTF 1.0 may be around for a while until 2.0 fully takes over,

andreasplesch:
[...]

PR #127, opened 2017-10-29
Binary MC for 1.0

andreasplesch:
I just noticed the `samples wanted` label in #109, so here we go.


1.0/genBinaryMC.sh
```diff
@@ -0,0 +1,17 @@
+# adjust according to actual installation path of gltf-pipeline
+alias gltfpipe="node /home/user/ehome/projects/gltf/gltf-pipeline-master/bin/gltf-pipeline.js"
```
emackey: Maybe use an npm dependency instead of a user-specific path here?

andreasplesch: I was following the style of the existing convertAll.sh script. 
Perhaps just remove the script to avoid confusion ?
Since I am unsure how to use an npm dependency here, I just commented the alias and added an explanation.

andreasplesch: The script is bit more complicated/strange because gltfpipe insisted on its 'output' directory.

andreasplesch:
If there is anything I can help with, I would be glad to.

emackey:
Hi @andreasplesch, sorry this fell through the cracks.  I don't think much/any effort is being put back into 1.0 now that glTF 2.0 is in full swing.  I don't think we need to include more permutations of 1.0 models here.  Also I'm mildly opposed to just automatically applying all permutations of extensions to all models; instead I think it would be more useful to have a specific example of each Khronos-ratified extension, showing how that extension is best used for some purpose that a core glTF model can't replicate.

andreasplesch:
https://github.com/cx20/gltf-test/issues/17
was the starting point for this PR since gltf-test would only accept gltfs from this repo.
At the time binary MC was the most compact encoding not requiring packaging of shaders.
If the glTF 2.0 MC extension stays in limbo, glTF 1.0 plus MC may be of more long-lived interest to some communities.
So, to me, being inclusive has more benefits than drawbacks. What is the concern ?

emackey:
I'll admit I haven't been following MC very closely, but I think the current work is being done in KhronosGroup/glTF#1163.

My concerns here are, in no particular order:

* We want to keep the spotlight on glTF 2.0, not 1.0.  New users might be diverted from 2.0 by seeing new models getting added to the 1.0 set.
* Sample models with textures take a lot of space relative to what Git is normally used to store, so space can become an issue.
* Having those models here implies Khronos is actively maintaining them, which is questionable now even for the existing 1.0 sample set, let alone new entries.  I suspect the existing set will only remain as a historical artifact, not an actively curated collection.

Sorry for raining on the parade.  But I think it's best to focus our efforts on making sure that 2.0 and its extensions can be made to cover use-cases that 1.0/MC used to cover.

(@cx20 feel free to chime in if you have opinions on this too, thanks!)

lexaknyazev:
JSYK, 1.0/MC has never been finished. 
I'm not even sure that existing draft implementations agree with each other.

andreasplesch:
Oh well, so be it.

PR #133, opened 2017-11-27
Add BoomBoxWithAxes model

bghgary:
See https://github.com/KhronosGroup/glTF/pull/1147.

I've also removed SmilingFace and WalkingLady from the README as the models have been removed and they are broken links right now.


pjcozzi:
Thanks @bghgary!

PR #134, opened 2017-11-30
Add a vertex colored model

msfeldstein:
None of the other models have vertex colors, so this is a simple cube with some colors in each vertex


cx20:
If possible, it is desirable that the file name and image size of the screenshot be the same as for other samples.
File Name : `screenshot/BoxVertexColors.png` -> `screenshot/screenshot.png` 
Image Size : `824x752` -> `128x128`

msfeldstein:
Thanks, updated

pjcozzi:
Thanks again @msfeldstein!

Does anyone else want to look at this before we merge?

cx20:
@msfeldstein I tried to display `BoxVertexColors.gltf` with several libraries.
However, the display result seems to be different from the screen shot. What kind of results do you expect to result? (Should alpha color be considered?)

BoxVertexColors/screnshot/screenshot.png
image

[Three.js r89dev + BoxVertexColors.gltf result:](https://cx20.github.io/gltf-test/examples/threejs/index.html?model=BoxVertexColors&scale=1)
image
```glsl
// Three.js Vertex Shader
#ifdef USE_COLOR
   attribute vec3 color;
#endif
```
[Babylon.js v3.1-beta-6 + BoxVertexColors.gltf result:](http://localhost:3000/examples/babylonjs/index.html?category=sampleModels&model=BoxVertexColors&scale=1&type=glTF)
image

```glsl
// Babylon.js Vertex Shader
#ifdef VERTEXCOLOR
  attribute vec4 color;
#endif
```

/cc @donmccurdy @bghgary

donmccurdy:
@cx20 in three.js >r88, you may want to add `renderer.gammaOutput=true` to test scenes. Many models will appear to dark without that. [three.js does not support alpha in vertex colors](https://github.com/mrdoob/three.js/issues/2118) at present.

cx20:
@donmccurdy Thanks for the advice. I understood that Three.js does not support alpha in vertex colors.
The test result when `renderer.gammaOutput=true` is set is as follows.
[Three.js r89dev + BoxVertexColors.gltf result:](https://cx20.github.io/gltf-test/examples/threejs/index.html?model=BoxVertexColors&scale=1)
image

msfeldstein:
yeah that last screenshot looks about right, it really just depends on what your lighting is like.  I can update it to show it in a well lit scene

msfeldstein:
Updated with a shot of what it looks like in don mccurdy's gltf viewer app

emackey:
Went looking for this yesterday and couldn't find it.  Didn't realize it was never merged, sorry about that!

csatobalint:
> None of the other models have vertex colors, so this is a simple cube with some colors in each vertex

Could you share the source code that generates the colored vertex box?

javagl:
@csatobalint Maybe the original creator can share more information here, but the glTF file says
`"generator": "FBX2glTF",`
and the PR contains the original model in FBX format. So apparently, the file https://github.com/msfeldstein/glTF-Sample-Models/blob/675d1b3435d8bfe1e4a47a0cf4b3d947a406e6a1/sourceModels/BoxVertexColors/BoxVertexColors.fbx has been converted to glTF with https://github.com/facebookincubator/FBX2glTF (and there is not really any "source code" involved... except for that of the conversion tool...)

emackey:
@csatobalint The [Blender glTF exporter](https://docs.blender.org/manual/en/latest/addons/import_export/scene_gltf2.html) also does vertex colors, and has a "Vertex Paint" mode.  But if you're looking to add vertex colors programmatically, there are various tools like cgltf, tinygltf, SharpGLTF, and others you can use.

bghgary:
We also have vertex color models here if it's useful:
https://github.com/KhronosGroup/glTF-Asset-Generator/tree/master/Output/Positive/Mesh_PrimitiveVertexColor

csatobalint:
I am trying to do it programatically in python. I have found this example: https://gitlab.com/dodgyville/pygltflib#create-a-mesh-convert-to-bytes-convert-back-to-mesh in which there nodes and triangles definitions by which an arbitrary shape can be generated. This example lacks the vertex color definitions and I could not figured out how to add that functionality. Can you or anybody help me with that? 
I have found earlier a matlab package that works similarly with vertex color definitions: https://www.mathworks.com/matlabcentral/fileexchange/109264-matlab2glb
The similar functionality is expected with this python package.

PR #137, opened 2017-12-01
Resize BoxTextured image to 256 so it is PowerOfTwo

msfeldstein:
The model currently has some issues on mobile as it uses a non power of two texture with a REPEAT wrap mode, which is not supported by mobile gpus.  While this is technically a valid gltf, it can cause lots of gotchas for people working on implementing a renderer, so maybe we should keep this example simple, and add a NonPowerOfTwo model for people to specifically be able to test this to implement the edge case


msfeldstein:
Oops didn't add the NPOT example

javagl:
Keeping the NPOT example is important. I'd even consider to omit the new NPOT example, and instead keep the original BoxTextured one with its NPOT textures: The point is that this is likely one of the first models that implementors will use for testing textures, and the "gotchas" that you mentioned should hit the implementor *as early as possible*. Otherwise, they could be tempted to say: "Fine, textures are working, next topic..." - and finally ship a renderer that *only* works with POT textures...

But that's just an opinion, let's see what others think.

2.0/BoxTexturedNonPowerOfTwo/README.md
```diff
@@ -0,0 +1,13 @@
+# Box Textured
+
+This model uses a Non Power-Of-Two texture with REPEAT mode wrapping.  This is an edge case that is technically a valid gltf model, but needs some renderer work to resize the texture before uploading to the GPU.  According to the spec:
+
+Non-Power-Of-Two Texture Implementation Note: glTF does not guarantee that a texture's dimensions are a power-of-two. At runtime, if a texture's width or height is not a power-of-two, the texture needs to be resized so its dimensions are powers-of-two if the sampler the texture references
```
lexaknyazev: The last statement is incomplete (compare with the spec). It may be better to put a link instead.

emackey:
There's a merge conflict here.  `MaterialsCommon` and `techniqueWebGL` were removed from master, so should be removed from this branch.

emackey:
Merge resolved.  Any other thoughts on this?  https://github.com/KhronosGroup/glTF-Sample-Models/pull/137#issuecomment-348487382

msfeldstein:
Up to you all. I think the standard basic box textured should be aPOT
example but I understand their reasoning for wanting to hit all edge cases
for example one. If that's what we want we can close this

emackey:
@javagl I think it's more in line with the other examples here, to keep the basic box simple, and call out features like NPOT with separate examples.  This way, engine developers can more easily isolate particular features for testing.  So, I think we should merge this.

javagl:
Sure, as long as there also is a non-POT-example. (Now there is a dedicated one. Otherwise, it could be part of one of the more general texture feature samples)

bghgary:
We will add some tests to [glTF-Asset-Generator](https://github.com/bghgary/glTF-Asset-Generator/issues/318) as a bonus.

emackey:
Merged.  Thanks again @msfeldstein and reviewers.

PR #140, opened 2017-12-11
Remove experimental extensions

### LINKED ISSUE #136: Remove unratified extension samples from master branch
Some 2.0 models have a samples using materials-common and webgl extensions (e.g. [Buggy](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/Buggy)). Both these extensions are not finalized. Having it in master is causing some confusion as to whether they can be used/tested against.

Should we consider removing them from master while they are still under development? They could be in another branch and move to master once they are done.

donmccurdy:
Fixes #136.

I've branched `master` to https://github.com/KhronosGroup/glTF-Sample-Models/tree/2.0-experimental-extensions, where the versions with experimental extensions are still available.


pjcozzi:
Thanks @donmccurdy!  Could you also add a note to the README.md that links to the `2.0-experimental-extensions` branch?  Could be a separate PR or just commit to master.

PR #141, opened 2017-12-23
Merge COLLADA animations.

### LINKED ISSUE #108: Merge animations in most samples, add sample using multiple animations correctly
I'm incredibly disappointed, I implemented 100% of the spec in Godot. Spent 4 days on it. At the time I'm done, I realize that, unlike what it proposes through the whole spec, glTF **does not** support multiple animations. 

Guys, It's 2017 and all DCCs support editing animation clips, so it should be perfectly possible to export all character animations in a single file without relying on cheap tricks like segmenting a timeline or having to export animations with the whole scene again in a separate file.

In the spec, the term **animations**, plural, is used through the whole thing. It totally makes sense, any animation works on a number of channels. By reading the spec it becomes obvious that multiple animations can be stored in the format.

Yet, when I go try the demos.. every demo I find uses an animation for a different node. For example on this demo:

https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/BoxAnimated

To get the animation to look like in the gif, you have to play all animations together.
In this other demo:

https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/CesiumMan

Each bone has an animation of it's own. you must play all animations together in order to see the guy walking.

So, please change the spec to make it very clear to implementers that, in 2017, a new format does not support multiple animations. Please also urgently add an official extension to support REAL multiple animations, so this format can be used for 21st century workflows.

### LINKED ISSUE #126: BoxAnimated clip durations do not match.
1. Previewing the BoxAnimated model here: https://gltf-viewer.donmccurdy.com/#model=https://raw.githubusercontent.com/KhronosGroup/glTF-Sample-Models/master/2.0/BoxAnimated/glTF/BoxAnimated.gltf
2. Open the "Animation" tab and press "playAll". (see https://github.com/KhronosGroup/glTF-Sample-Models/issues/108)

After pressing `playAll`, the two clips (position and rotation) line up for the first runthrough, but their durations are different, so after this they stop lining up. Same issue occurs in the BabylonJS sandbox. In the console I find:

screen shot 2017-10-27 at 8 32 23 am

I believe this has the same underlying cause as #108, namely that these two clips should have been a single `animations` entry, inherently with the same length, but do not.

donmccurdy:
Regenerated models after https://github.com/KhronosGroup/COLLADA2GLTF/pull/102.

Fixes #126, fixes #108.


pjcozzi:
Thanks @donmccurdy!

@reduz do you want to check out these newly generated models per your comments in #108?

bghgary:
I did a quick scan with a tweaked BabylonJS sandbox (to play only the first animation) and this change seems good to me.

pjcozzi:
Thanks @bghgary @donmccurdy!

PR #143, opened 2018-01-08
New model testing out vertex colors in COLOR_0

emackey:
This model tests out vertex colors in your rendering engine.  If it works, it looks like this:

screenshot

Really I built this just to review AnalyticalGraphicsInc/cesium#6089.  Vertex colors are being added to Cesium, but not available in master yet as of this writing.  This is the kind of thing you see when vertex colors are not implemented:

sample_fail

There are "X" marks in the "wrong" color channels, and the X's can only be hidden when the vertex colors are correctly applied.


pjcozzi:
Looks cool, anyone want to review this?

donmccurdy:
Passes validation, works in three.js, and here's the ZIP version:

[Uploading VertexColorTest.zip…]()

Looks good to me.

cx20:
I confirmed that this model can be displayed with 3 libraries.

image
BTW, Hilo3d can display BoxVertexColors.gltf, but it does not pass the test of VertexColorTest.gltf. Is it a problem with the library?

emackey:
> Hilo3d can display BoxVertexColors.gltf, but it does not pass the test of VertexColorTest.gltf. Is it a problem with the library?

The main difference is that VertexColorTest.gltf also uses a mesh and a material *without* vertex colors, in the same model as one that does use vertex colors.  This combination revealed a bug in Cesium's upcoming implementation which has since been fixed (but not yet released, 1.42 should have vertex color support).

I don't know if that's the same problem that Hilo3d has.  The other difference is that this model is testing `VEC3` colors while the other model tests `VEC4` colors, that could be a factor too.

cx20:
I confirmed that it has been fixed by the latest version of Hilo3d.
As for Cesium, I would like to test it again when a new version comes out.

pjcozzi:
@cx20 I believe Cesium will be fixed in Cesium 1.42 on Feb 1.

@lilleyse can you confirm?

lilleyse:
Yes it will be in Cesium 1.42.

cx20:
I confirmed that the test model will be displayed in Cesium 1.42.
image

pjcozzi:
@cx20 cool, thanks for testing so promptly!

PR #116, opened 2017-09-07
Added sample models for Draco_mesh_compression_extension

fanzhanggoogle:
Hi,

This PR adds sample models for usage of [KHR_draco_geometry_compression](https://github.com/KhronosGroup/glTF/pull/874). These models are created using [COLLADA2GLTF](https://github.com/KhronosGroup/COLLADA2GLTF/pull/79). Please see the `glTF-Draco` folder in each model folder. The models could be loaded by modified [ThreeJS GLTF loader](https://github.com/fanzhanggoogle/gltf-test/tree/draco_2.0_1.0/libs/three.js/draco).

Here's preliminary compression result on the meshes of sample models:
### Comparison on geometries (mesh/primitive only)
| Model Name | Total Geometry Size (byte) | Compressed Size (byte) | Compression Ratio |
| ------------- | ------------- | ------------- | ------------- |
| 2CylinderEngine  | 1794612 | 183530 | 9.778303275 |
| Avocado | 23580  | 3988 | 5.912738215 | 
| BarramundiFish | 128208  | 17448 | 7.348005502 | 
| BoomBox | 207816  | 27651 | 7.515677552 | 
| Box | 648  | 160 | 4.05 | 
| BrainStem | 2009628 | 167491 | 11.9984 | 
| Buggy  | 7745988  | 692824 | 11.18031131 | 
| CesiumMan  | 211320  | 17535 | 12.05132592 | 
| CesiumMilkTruck  | 108144  | 14785 | 7.31444 | 
| Corset | 662184  | 83635 | 7.917546482 | 
| Duck  | 102040  | 11354 | 8.987141096 | 
| GearboxAssy  | 4915704  | 423307 | 11.61262157 | 
| Lantern | 231324  | 28535 | 8.106676012 | 
| Monster  | 48984  | 6664 | 7.350540216 | 
| ReciprocatingSaw  | 3496764 | 333415 | 10.48772251 | 
| RiggedFigure  | 19296  | 2830 | 6.81837 | 
| RiggedSimple  | 5736  | 772 | 7.430051813 | 
| VC  | 669786  | 152188 | 4.401043 | 
| WaterBottle | 149412  | 15146 | 9.864782781 | 

### Comparison on glTF files (including texture, gltf file and other data)
| Model Name | Uncompressed GLB File Size  (byte) | Compressed GLB File Size (byte) | Compression Ratio |
| ------------- | ------------- | ------------- | ------------- |
[...]

PR #145, opened 2018-01-26
Update Draco gltf models using ratified bitstream

FrankGalligan:
Models were created from branch:
https://github.com/google/draco/tree/gltf_2.0_draco_extension


FrankGalligan:
@donmccurdy Can you try these files? I might redo them with more compression later. But I just want to make sure they work for you.

donmccurdy:
These are working with https://github.com/donmccurdy/three.js/pull/6, thanks 👍

donmccurdy:
In [./javascript](https://github.com/google/draco/tree/gltf_2.0_draco_extension/javascript), should there be a glTF flavour of the encoder? Or are they the same?

lexaknyazev:
@FrankGalligan 
Why do some attributes in GearboxAssy model point to an accessor with indices?
https://github.com/FrankGalligan/glTF-Sample-Models/blob/dae68f182050542e0551172ec8178b739493ff6b/2.0/GearboxAssy/glTF-Draco/GearboxAssy.gltf#L2431-L2436

This causes validation errors:
https://travis-ci.org/KhronosGroup/glTF-Sample-Models/builds/333954339#L883

FrankGalligan:
@lexaknyazev filed https://github.com/FrankGalligan/gltf-pipeline/issues/1

pjcozzi:
What's left before we can merge this?

lexaknyazev:
How does this PR relate to #116?

donmccurdy:
> How does this PR relate to #116?

Those models were created with a pre-spec version of the encoder. I believe this PR can replace #116.

donmccurdy:
> What's left before we can merge this?

I've tested 4-5 of these in the three.js Draco implementation, and they looked good to me.

pjcozzi:
Some of of these - and soon all - have also been tested with Cesium.  CC https://github.com/AnalyticalGraphicsInc/cesium/pull/6191

bghgary:
The fallback attributes are not working for me. Looks like `bufferView` is missing from the accessors.

https://github.com/FrankGalligan/glTF-Sample-Models/blob/be470ea72ecb7adf2b30932349d1b12164af07b7/2.0/Box/glTF-Draco/Box.gltf#L47

https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/schema/accessor.schema.json#L11

bghgary:
I just noticed that extensionsRequired is set for these. If the fallback buffers are intended to work, then this should be removed. If not, then the fallback buffers should be removed.

donmccurdy:
I believe accessors with missing bufferviews is the expected Draco-required output.

donmccurdy:
From the spec: 

> Then the loader must process `attributes` and `indices` properties of the `primitive`. When loading each `accessor`, you must ignore the `bufferView` and go to the previously decoded Draco geometry in the `primitive` to get the data of indices and attributes.

We implemented it this way in three.js:

```js
if ( accessorDef.bufferView === undefined && accessorDef.sparse === undefined ) {

  // Ignore empty accessors, which may be used to declare runtime
  // information about attributes coming from another source (e.g. Draco
  // compression extension).
  return null;

}
```

/cc @lexaknyazev

bghgary:
Yes, sorry for my confusion. I reread the spec and I now understand what it is supposed to look like when the extension is required.

I think we can add a couple of minor things to avoid confusion.
* Add an short explanation to the extension required part in the spec and indicate that the only difference is that `buffer` is missing from the `bufferView`.
* Add a sample model of Draco with a fallback where Draco is only in `extensionsUsed`. We can also add this to the asset generator project instead if that's better.

donmccurdy:
> Add an short explanation to the extension required part in the spec and indicate that the only difference is that bufferView is missing from the accessor.

+1 for adding the explanation. Let's do that after the PR is merged.

> Add a sample model of Draco with a fallback where Draco is only in extensionsUsed. We can also add this to the asset generator project instead if that's better.

Also +1. I tried to keep this case in mind with three.js implementation, but haven't tested it and would be surprised if it works yet.

pjcozzi:
@bghgary @donmccurdy submitted https://github.com/KhronosGroup/glTF/issues/1244.

Please hit merge if this is otherwise ready. :rocket: :rocket: :rocket:

bghgary:
I tried all the models with a glTF-Draco folder. All of them are working for me except:
* CesiumMan
* Monster
* RiggedFigure

Does this work for anyone else? It would be good to know if it's the model before I dig in to why.

donmccurdy:
Those three are working for me, yes. See: https://github.com/mrdoob/three.js/pull/13194

bghgary:
I think I figured it out, but would like confirmation. It looks like the unique ids are mapped wrong in RiggedFigure. I haven't checked the rest of them.

bghgary:
The JOINTS_0 attribute is returning floats that are not integers.

```
0: 2.000000238418579
1: 3.0021979808807373
2: 4.000000476837158
3: 0
```

Is this expected?

bghgary:
I fixed this issue from my end and all of the models are passing for me now. I am okay with merging if the JOINTS_0 issue is expected.

donmccurdy:
Confirmed *Rigged Figure* is working in three.js, but as Gary mentions the joint indices are floats not ints yes.

FrankGalligan:
I think the spec is fine as it should be int. Even the gltf for the draco version says the data should be int:
https://github.com/FrankGalligan/glTF-Sample-Models/blob/be470ea72ecb7adf2b30932349d1b12164af07b7/2.0/RiggedFigure/glTF-Draco/RiggedFigure.gltf#L2481

I think this is an implementation bug.

donmccurdy:
Thanks all! Merging this, and we can address the JOINTS_0 implementation bug with a followup. Meanwhile, opened another issue here: https://github.com/KhronosGroup/glTF-Sample-Models/issues/147

PR #148, opened 2018-03-01
Update the TextureCoordinateTest model

### LINKED ISSUE #118: TextureCoordinateTest models use extreme materials
This is a nitpick in the larger scheme of things, but the pbrMetallicRoughness structs in the TextureCoordinateTest model specify neither metallicFactor nor roughnessFactor, ending up with the default 100% for each, which is a very light-averse material that's likely to show up pitch black in some less-carefully-configured viewers. Since the stated purpose of this model is to help developers figure out what they're doing, perhaps at early stages of development, I think it might be nice to specify a material that reflects a decent amount of both diffuse and direct lights.

### LINKED ISSUE #146: TextureCoordinateTest still using -z forward?
//cc @emackey looks like the texturecoordinatetest sample might still be using -z forward and needs to be updated to +z?

emackey:
* Fixes #118 material settings.
* Fixes #146 by switching to +Z forward.

@bghgary @sbtron The `.glb` file in this PR fails to load in Mixed Reality Viewer, but passes validation with no messages.  Is there some diagnostic tool I can get that would tell me why the viewer is rejecting it?


bghgary:
> The .glb file in this PR fails to load in Mixed Reality Viewer, but passes validation with no messages. Is there some diagnostic tool I can get that would tell me why the viewer is rejecting it?

Not for the end user no, but the latest version of the app does load it. There were bugs before that have since been fixed. Can you check what version of MRV you are on (Menu -> About)?

emackey:
Hmm, can't copy/paste the version number, had to type it:  2.1801.4012.0

I'm not on any insider build here, patches are managed by the local IS department.

bghgary:
Thanks, that build is older than mine and mine works. Someone else reported a similar issue on twitter to me earlier this week. In any case, the bug has been fixed and will eventually make it to users.

emackey:
Great.  This should be good to merge then.

PR #150, opened 2018-03-02
3 d viewer beta

AsoboStudio:
Low poly fox by PixelMannen (Animated)
https://sketchfab.com/models/371dea88d7e04a76af5763f2a36866bc
https://opengameart.org/content/fox-and-shiba

Copyright/Attribution Notice:
Feel free to credit me. It is not required though.


emackey:
Nice model!

I notice it has what appear to be a number of animation cycles, walk, run, look around, etc.  But these are all lumped together into a single animation block called `Take 001`.  Do you have any way to break these up into separate named animations?

This repo is in need of models that have separate, named animations, such that for example the run cycle could be used independently of a walk cycle.

Also, is there a license associated with this?

AsoboStudio:
Yes, this file contains only one animation; and I wouldn't try to break this into seperate named animations.
Other GLB files may contain multiple named animations.
As I understand from the above links, the author gives us full rights to use his asset.

donmccurdy:
I love this model, looks great. 

Will have to agree with @emackey though, that we probably should not show it as a sample unless the animations are split apart and looping cleanly. There has been [some confusion](https://github.com/KhronosGroup/glTF/issues/1052) about whether and how glTF 2.0 supports multiple animations because our sample models were misleading in the past, and we don't want to repeat that issue. 🙂

I've been working on [a three.js feature](https://github.com/mrdoob/three.js/pull/13430) that would let us split the clip and re-export, and will test that with this model. But unfortunately there's no guarantee it will work once separated, unless the asset was animated with that case in mind.

donmccurdy:
Unfortunately i'm not getting clean results when trying to split the animation at runtime in three.js (which isn't necessarily surprising if the model wasn't exported with that intent), and importing into Blender doesn't quite look right either, so not obvious to me how to split the animations from the FBX source.

scurest:
No interest in this?

I took the GLB down and polished it quite hard to get this glTF-Embedded file: https://gist.github.com/scurest/1c0f5fb076c3ff128bde2dd6dcb3a796.

It has three separate animation cycles, plus several other optimizations. It is much smaller than the original GLB too.

How's it look?

donmccurdy:
Looks good! Quite compact, too. Couple things:

- [ ] the original asset is **CC Attribution**, we should include a link to the author's sketchfab page.
- [ ] looks like there is an extra frame or two on the run cycle, it doesn't quite loop cleanly. that's there in the sketchfab version too, though... would it be difficult to fix that?

Otherwise I'm happy with this, it passes validation and we do need more samples with multiple animations. 👍

scurest:
* Good point. How about an `asset["copyright"]` string like

    ````
    Original model by PixelMannen <https://opengameart.org/content/fox-and-shiba>. Rigging and animation by Tom Kranis <https://sketchfab.com/models/371dea88d7e04a76af5763f2a36866bc>, CC Attribution <https://creativecommons.org/licenses/by/4.0/>. Modified for use as a glTF sample model.
    ````

    I think that covers CC Attribution's requirements?

* The "catch" when the left forepaw strikes the ground? Like you say, that was in the original, and removing it doesn't seem to be a matter of just deleting keyframes, so I think that's beyond my skill, unfortunately.

donmccurdy:
For copyright, I think that and including a `README.md` with the sample file here would be sufficient.

Would defer to others on the catch in the animation loop; haven't had the chance to try editing that yet myself.

scurest:
I revised the run animation a bit. Any better?

donmccurdy:
Thanks! I like the new version. 👍 Do you, or @AsoboStudio, want to open a PR (or update this one) including a `README.md` file?

emackey:
Further discussion moved to new PR #249.

PR #152, opened 2018-03-13
Added Orientation Test model

emackey:
screenshot

This was created to test node orientation problems found in an up-and-coming glTF rendering engine.

/cc @abwood


emackey:
https://twitter.com/pjcozzi/status/973928201876656128  😄

cx20:
I tried this model with [gltf-test](https://github.com/cx20/gltf-test). The current status is as follows.
image
If I do not misunderstand the purpose of the test, every library seems to be displayed correctly.
I thought it would be nice to have an example (screen) of how it would appear if there was a problem with rotation.

BTW, [Khronos glTF Loader](https://github.com/KhronosGroup/glTF-WebGL-PBR/) has failed to load models, I think that it is a different issue from the problem of rotation.

emackey:
> If I do not misunderstand the purpose of the test, every library seems to be displayed correctly.

Yes, this is one of the first things a new glTF implementation needs to get right, so it's not a surprise that our regularly-tested set is doing fine with this model.

> I thought it would be nice to have an example (screen) of how it would appear if there was a problem with rotation.

Added to the README, thanks for the suggestion.

orientationtestfail

You can see in the fail image that the arrows point away from their targets.  The targets are baked into place, they do not need any node transformations applied to be correct.  The arrows that point at those targets, however, are axis-aligned in their native orientations, so depend on correct processing of node orientations and transformations to be able to point at the targets.

PR #159, opened 2018-03-31
Update flight helmet

bghgary:
* Fixes issue with extra UV sets in the asset
* Collapse a bunch of the meshes reducing draw calls


emackey:
Odd that "baseColor1" and "baseColor2" type textures all traded places.  But the new model looks good and passes the validator's new tests for unused texture coordinates.

bghgary:
Yeah, I'm not sure sure why it did that, but it shouldn't matter.

emackey:
Do the materials have names?  Maybe the tool could use those instead of the index numbering.  No need to re-do existing models like this one, but, users may appreciate the consistency of output.  Just something to consider.

PR #167, opened 2018-05-03
Add feature test model for KHR_texture_transform

bghgary:
The extension is not ratified yet, so I'm not sure we should merge this until it is. I also wanted to include a link to the spec, but I didn't want to link to the not yet merged spec. We should probably at least wait until [the PR](https://github.com/KhronosGroup/glTF/pull/1015) is merged.


bghgary:
Also, do we expect people have to manually run generate-index.py when adding a new model?

cx20:
I added the current PR model to [gltf-test](https://github.com/cx20/gltf-test#feature-test-models) and tested it.
https://github.com/cx20/gltf-test#feature-test-models
image
Since specifications are not fixed yet, I plan to test again when there is an update.

emackey:
> do we expect people to manually run generate-index.py when adding a new model?

I plan to re-run this as needed.  But feel free to run it if you have Python already installed and can get it to run (it's not yet compatible with Python 3, but this too will be fixed).

2.0/README.md
```diff
@@ -62,14 +62,15 @@
...
+| [Normal Tangent Mirror Test](NormalTangentMirrorTest) | ![](NormalTangentMirrorTest/screenshot/screenshot.png)        | Tests an engine's ability to load supplied tangent vectors for a normal map. |
+| [Orientation Test](OrientationTest)                   | ![](OrientationTest/screenshot/screenshot.png)                | Tests node translations and rotations. |
+| [Texture Coordinate Test](TextureCoordinateTest)      | ![](TextureCoordinateTest/screenshot/screenshot.png)          | Shows how XYZ and UV positions relate to displayed geometry. |
+| [Texture Settings Test](TextureSettingsTest)          | ![](TextureSettingsTest/screenshot/screenshot.png)            | Tests single/double-sided and various texturing modes. |
+| [Vertex Color Test](VertexColorTest)                  | ![](VertexColorTest/screenshot/screenshot.png)                | Tests if vertex colors are supported. |
+| [Texture Transform Test](TextureTransformTest)        | ![](TextureTransformTest/screenshot/screenshot.jpg)           | Tests if KHR_texture_transform extension is supported properly. |
```
emackey: I wonder if we should make a separate section for "Extension Test Models"?  Or do we expect all the major implementations to eventually support all the KHR extensions?

2.0/TextureTransformTest/README.md
```diff
@@ -0,0 +1,37 @@
+# Texture Coordinate Test
```
bghgary: Oops, copy and paste error here. Will fix.

bghgary:
I have fixed the READMEs.

> I wonder if we should make a separate section for "Extension Test Models"?

I did this, but there is still a bit of inconsistency. All of the other extensions go in a differently named folder. For example, the Box model has `glTF` and `glTF-pbrSpecularGlossiness`. This test model doesn't fit this pattern. I'm personally okay with this.

> Or do we expect all the major implementations to eventually support all the KHR extensions?

We intend on implementing all possible KHR extensions in BabylonJS. Not sure about others.

emackey:
Thanks @bghgary!

emackey:
Updated model-index.json in 7f3a2ee8b5ab7e474.

lexaknyazev:
Could we have an additional sample model for this extension with texture transform applied to SpecGloss material to test engines against "extension-within-extension" case?

bghgary:
Yes, I will look into it.

PR #175, opened 2018-06-11
Add the Sponza model.

Themaister:


pjcozzi:
Very nice, thanks @Themaister!

Does anyone want to quickly review?

lexaknyazev:
I'm afraid this was merged too soon.
- Why GLB format is used despite all images are stored externally?
- Could validation issues be fixed (they're not critical, so CI is green)?
```json
[
            {
                "code": "MESH_PRIMITIVE_UNUSED_TEXCOORD",
                "message": "Material does not use texture coordinates sets with indices (0).",
                "severity": 2,
                "pointer": "/meshes/0/primitives/50/material"
            },
            {
                "code": "UNUSED_OBJECT",
                "message": "This object may be unused.",
                "severity": 3,
                "pointer": "/materials/3"
            }
]
```

Themaister:
That is probably caused by some objects in the OBJ which were left unused.
As for glB, that's what my exporter uses. It embeds the vertex attributes without having to resort to base64, so that's something. Is it expected that GLB files have baked in all textures?

lexaknyazev:
It's just a convention of this repo: 
- `glTF` folder contains all files separately (`.gltf`, `.bin`, images).
- `glTF-Binary` keeps everything in one `.glb` file.

lexaknyazev:
Data URI (base64) embedding is very inefficient to store and to load, so it shouldn't be used for assets of several megs size.

pjcozzi:
My bad, @lexaknyazev. :face_with_head_bandage: 

Looks like @Themaister is already on it. :medal_sports:

cx20:
@lexaknyazev  I think it would be better to follow the convention. However, since this sample is very large (over 150 MB), we are afraid that if all the patterns are recorded, the repository will become enlarged.

cx20:
It is better to match the file name with the folder name. From sponza.glb to Sponza.glb.

emackey:
Several images in this 150 MB model could be dramatically reduced in size.  For example `7385312956576678639.png` has information only in the green channel, but supplies a zeroed-out alpha channel, increasing both the size and the difficulty in working with the image directly (we expect new users to examine all parts of the assets, so ease of loading into paint programs is desirable when possible in this official sample repo).  Many of these images could be converted to JPG format to save space, after having their alpha channels discarded.

Also, we should try to stick to the conventions of this repository.  It might be best to keep the images separate but convert the .glb to .gltf.  I think every sample model has a basic .gltf form, and all of the other forms are optional.

lexaknyazev:
Since this has been merged, any further action without using force-push would only increase the repo size.

06wj:
I rebase my [PR](#176) to 395358db39e9310bde16aeb65bbea5e32d184d47, it's OK to use force-push.

Themaister:
I've force-updated my branch: https://github.com/Themaister/glTF-Sample-Models/commit/ba1cc7515efb854ad93027ea15715dd15176c7ee

- Use JPG for non-transparent textures
- Change to glTF format with separate .bin and textures
- Fix the warnings.
- Hopefully I didn't screw the model, but looks okay to me.

cx20:
@Themaister  I confirmed that your branch's model is downsized to 50 MB in size.

PR #176, opened 2018-06-13
Add Multiple UV Test model

06wj:
screenshot
This model has two uv texture coordinates.


donmccurdy:
Looks good! This model works in BabylonJS but (as expected) appears wrong in three.js which does not yet support multiple UVs fully. I'm OK with merging it despite that, it is a good test case. :)

06wj:
@emackey Can this be merged to the ```master``` branch?

emackey:
Thanks for your patience @06wj.  Can we show something better than a broken glTF logo when an engine doesn't support multi UVs?  Maybe an image of the message:

```
Multiple UVs
not supported
in this viewer
```

Or something along those lines?

lexaknyazev:
We should also remove two empty nodes (with camera and lamp).

06wj:
@emackey  Update the model.
When engine doesn't support multi UVs, it will show the error message.

lexaknyazev:
Not a strong opinion. why do we need an asset-provided camera for this sample? Is it aligned with other models in this repo?

emackey:
> why do we need an asset-provided camera for this sample?

Not strictly needed but I don't think it hurts anything by its presence.  It doesn't appear to produce a validation message anymore.

emackey:
No strong objections were reported here, so I'm merging this.  Let me know if anyone finds any issues.

PR #183, opened 2018-07-07
Fixing some issues with the PBR models

### LINKED ISSUE #179: Waterbottle.glb is incorrect
The gltf-binary version waterbottle is incorrect and has different material textures compared to the rest of the variations like glTF etc.

bghgary:
Fixes #179

* Added the source for the FlightHelmet model
* Use PNG instead of PSD for the WaterBottle since newer Unity builds are having issues with the PSDs
* Found an issue with my Unity exporter that didn't handle Unity's sRGB flag properly which affects occlusion and metallic textures


emackey:
Thanks @bghgary.

Issue #181, opened 2018-07-05
SpecGloss samples


lexaknyazev:
It would be nice to have a sample model showing SpecGloss parameters like MetalRoughSpheres.

Also, it would be useful to have an asset with two identical models showing the same visuals with MR and SpecGloss.

bghgary:
> Also, it would be useful to have an asset with two identical models showing the same visuals with MR and SpecGloss.

Are you asking for any models that do this or specifically MetalRoughSpheres? If it's the former, all the SpecGloss models that I put up already do this (e.g. https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/BoomBox)

lexaknyazev:
I'm asking for an asset that features two instances of the same geometry. E.g. a scene containing two boomboxes that look alike but the first one uses MR while the second uses SpecGloss.

emackey:
> I'm asking for an asset that features two instances of the same geometry [...] that look alike but the first one uses MR while the second uses SpecGloss.

@lexaknyazev This is a (relatively) easy hand-edit to any of the existing specGloss models that contain metalRough fallback textures.  Just copy/paste the mesh and material, and edit one to require spec/gloss and the other to use only MR.  Then use an X translation to separate the two meshes.

For example, here's the edit applied to the WaterBottle model:  [TwoWaterBottles.zip](https://github.com/KhronosGroup/glTF-Sample-Models/files/2181008/TwoWaterBottles.zip)  Extract this into the `2.0\WaterBottle\glTF-pbrSpecularGlossiness` folder.  You may diff it against the original spec/gloss WaterBottle.gltf to see what I did to it.

@donmccurdy I noticed while doing this that ThreeJS treats the spec/gloss 1.0-glossiness very differently from the metal/rough roughness.  I don't think it's a problem with this model as BabylonJS shows a much closer match.  I know there was a recent fix to roughness values in ThreeJS, is it possible the fix was not applied to glossiness values?

![twowaterbottlesinthreejs](https://user-images.githubusercontent.com/1235080/42520657-c3cc2002-8434-11e8-8125-564861c0b043.png)

lexaknyazev:
Looks like this was a useful idea. In such case how about we update all SpecGloss models to the following form?
- an instance with only MR;
- an instance with only SpecGloss;
- an instance with both.

donmccurdy:
[...]

PR #192, opened 2018-09-06
Spec gloss vs metal rough

emackey:
This adds the "two water bottles" side-by-side comparison of spec/gloss and metal/rough as a new sample model in its own right.  New labels were added, and they too are part of the test.

/cc #181 and AnalyticalGraphicsInc/cesium#7006

/cc @bghgary @OmarShehata @abwood


emackey:
See the [README](https://github.com/KhronosGroup/glTF-Sample-Models/blob/f72fc12a44323f0d65f5220de2426bff4abef116/2.0/SpecGlossVsMetalRough/README.md)

lexaknyazev:
"License Information" part of the readme contains only authors, not usage / distribution license.

emackey:
Yes, this is because it is derived from the Microsoft WaterBottle, which similarly lists authors but not a license.  @bghgary or @sbtron can a real license be added to WaterBottle and friends?  Preferably one that permits derivative works!

bghgary:
I'll let Patrick Ryan answer this in more detail. I think we wanted to avoid putting any license because it still caused restrictions even with the most permissive one. We wanted this to be completely open for use. @PatrickRyanMS what do you think?

lexaknyazev:
Lack of license usually means that nothing could be done with the model without explicit permission.
To give it away completely, the typical way is to declare it "public domain" (if applicable) or use something like CC0.

PatrickRyanMS:
I agree with @bghgary that we want these assets publicly distributable and open for anyone to use, remix, share, etc. The issue with even applying a CC-BY 4.0 license is that since we work for Microsoft and these contributions are part of our work for the company, I don't know that we can just assign a license, even CC-BY, without getting our legal department to sign off on us doing that. This is a question that we will need to dig a little further on before we can attach any legal language to the assets, but we do want them to be public domain.

emackey:
@bghgary @PatrickRyanMS Thanks, and please do follow up on this.  @lexaknyazev is correct that legally we shouldn't be doing anything with the Microsoft-supplied models, not even including them for distribution with the sample models, without a license.

cx20:
@emackey I added this model to [gltf-test](https://github.com/cx20/gltf-test).

I think the current status is as follows.
image

| ||
|:------:|:------:|
|[image](https://cx20.github.io/gltf-test/examples/threejs/index.html?category=tutorialModels&model=SpecGlossVsMetalRough&scale=10&type=glTF)|[image](https://cx20.github.io/gltf-test/examples/babylonjs/index.html?category=tutorialModels&model=SpecGlossVsMetalRough&scale=10&type=glTF)|
|[image](https://cx20.github.io/gltf-test/examples/xeogl/index.html?category=tutorialModels&model=SpecGlossVsMetalRough&scale=10&type=glTF)|[image](https://cx20.github.io/gltf-test/examples/Hilo3d/index.html?category=tutorialModels&model=SpecGlossVsMetalRough&scale=10&type=glTF)|
|[image](https://cx20.github.io/gltf-test/examples/playcanvas/index.html?category=tutorialModels&model=SpecGlossVsMetalRough&scale=10&type=glTF)|[image](https://cx20.github.io/gltf-test/examples/czpg/index.html?category=tutorialModels&model=SpecGlossVsMetalRough&scale=10&type=glTF)|

emackey:
@cx20 Thanks.  I didn't realize the spec/gloss ecosystem was in such bad shape.  Cesium support is coming soon.  Some of the others appear to be really close, like ThreeJS and Hilo3D.

emackey:
@bghgary @PatrickRyanMS How should we proceed with this one?  Did you reach out to legal, or is there any other option?

bghgary:
@sbtron is going over this with legal which takes some time. @sbtron, any status on this?

emackey:
OK, in the master branch, Microsoft has added CC0 licenses to its sample models (thanks so much @bghgary and @sbtron!).  I've copied that license here to this new model, since it's just a re-packaging of two different versions of the existing water bottle model side-by-side.

This should be good to merge now.

2.0/SpecGlossVsMetalRough/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+No fallback to metal/rough is provided for the spec/gloss side, so unlike most other samples, `KHR_materials_pbrSpecularGlossiness` extension support is explicitly required by this model.  The model is expected to fail to load on viewers that do not support the extension, as such support is not required by the core glTF specification.  This model marks the extension under `extensionsRequired`, meaning any viewer that doesn't support the extension should reject the model completely.  So, a complete failure to load the model, possibly with a message about lack of support for the extension, is not considered a test failure for core glTF 2.0.
+
+However, if the model loads into a viewer, and the two water bottles don't look almost exactly the same, then that does constitute a test failure.
+
+The labels floating above the water bottles should also have the same visual appearance (other than the text of the label being different).  One should not be shiny or more reflective than the other.
+
+## License Information
+
+Based on the Water Bottle sample model, donated by Microsoft for glTF testing.
```
bghgary: I don't think it's appropriate to say "donated by Microsoft for glTF testing" here?

2.0/SpecGlossVsMetalRough/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+[![CC0](http://i.creativecommons.org/p/zero/1.0/88x31.png)](http://creativecommons.org/publicdomain/zero/1.0/)
```
bghgary: Add two spaces at the end of the line to put the text on a separate line.

2.0/SpecGlossVsMetalRough/README.md
emackey: Removed.

2.0/SpecGlossVsMetalRough/README.md
emackey: Fixed.  I had copy/pasted this, but I have my copy of VSCode configured to remove trailing spaces automatically.  Sad that markdown places significance on such things.

cx20:
@emackey @donmccurdy I have doubts about the display result of Three.js.
The Metallic Roughness model reflects the environment map, but the Specular Glossiness model does not seem to reflect the environment map. Is this a correct display result?

[Three.js + SpecGlossVsMetalRough.gltf result:](https://cx20.github.io/gltf-test/examples/threejs/index.html?category=tutorialModels&model=SpecGlossVsMetalRough&scale=10&type=glTF)
image

donmccurdy:
Thanks again for these side-by-side comparisons. ❤️

It's entirely possible the spec/gloss result is incorrect in three.js, I'll file a bug.

PR #194, opened 2018-09-24
Add link to "30+ pirate themed (public domain) models"

pjcozzi:

PR #195, opened 2018-09-24
Tweak pirate model link in README

emackey:
Replace the reddit link with the asset download link, and make it a sub-bullet under "Kenney Assets" which was already in place above.


emackey:
For #194

emackey:
Added a section on model sharing services: Sketchfab (moved link from below), Remix3D, and Poly.

PR #196, opened 2018-09-26
Added animated morph cube COLLADA model

lasalvavida:
Part of testing https://github.com/KhronosGroup/COLLADA2GLTF/pull/206

The geometry and animations should match exactly the glTF 2.0 model that we have and I did my best to get the names to match where possible when run through COLLADA2GLTF.

The transform decomposition and material mappings are a little lossy (since COLLADA doesn't have PBR), but it should be a pretty close match.


emackey:
Thanks @lasalvavida.

PR #199, opened 2018-10-03
Added hints about the requirements for contributions

javagl:
- Updated the link to show the glTF 2.0 directory structure
- Added hint that a screenshot is desirable
- Mentioned that the license information should be included

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,8 +41,12 @@ We appreciate sample model contributions; they help ensure a consistent glTF eco
 
 To contribute a model, open a pull request with:
 * A new subdirectory containing
-   * The model in as many glTF variations as possible (using the same directory structure as the others ([example](1.0/Box))).  Tools for converting to glTF are [here](https://github.com/KhronosGroup/glTF#converters).
-   * A README.md with any usage restrictions.
+   * The model in as many glTF variations as possible (using the same directory structure as the others ([example](2.0/Box))).  Tools for converting to glTF are [here](https://github.com/KhronosGroup/glTF#converters).
+   * A screenshot of the model, stored in a subdirectory called `/screenshot`
+   * A README.md with information about the model. As shown in [this example](https://raw.githubusercontent.com/javagl/glTF-Sample-Models/master/2.0/Box/README.md), this file should at least include the following elements:
+      * The name or title of the model
+      * An embedding of the screenshot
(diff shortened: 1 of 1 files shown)


javagl:
As discussed during the previous call: Contributors should include license information for their models.

I also added a hint that the README.md should include a screenshot, because this is currently the case for basically *all* models. If there are objections, I can remove this part, but it seems reasonable to mention this, considering that it will basically be required for the overview table.

README.md
```diff
@@ -41,8 +41,12 @@ We appreciate sample model contributions; they help ensure a consistent glTF eco
 
 To contribute a model, open a pull request with:
 * A new subdirectory containing
-   * The model in as many glTF variations as possible (using the same directory structure as the others ([example](1.0/Box))).  Tools for converting to glTF are [here](https://github.com/KhronosGroup/glTF#converters).
-   * A README.md with any usage restrictions.
+   * The model in as many glTF variations as possible (using the same directory structure as the others ([example](2.0/Box))).  Tools for converting to glTF are [here](https://github.com/KhronosGroup/glTF#converters).
+   * A screenshot of the model, stored in a subdirectory called `/screenshot`
+   * A README.md with information about the model. As shown in [this example](https://raw.githubusercontent.com/javagl/glTF-Sample-Models/master/2.0/Box/README.md), this file should at least include the following elements:
```
lexaknyazev: Could a link to `README.md` be relative here?

javagl: The intention was to make it a *raw* link, so that people can easily Copy+Paste the contents of the README.md that it links to (e.g. the image link, with this magic `!`), without having to manually go to the "raw" view. If it is OK that users will have to manually select the "raw" view, I can change it.

lexaknyazev: I see. In such case, could the link point to the original KhronosGroup repo?

README.md
```diff
@@ -41,8 +41,12 @@ We appreciate sample model contributions; they help ensure a consistent glTF eco
 
 To contribute a model, open a pull request with:
 * A new subdirectory containing
-   * The model in as many glTF variations as possible (using the same directory structure as the others ([example](1.0/Box))).  Tools for converting to glTF are [here](https://github.com/KhronosGroup/glTF#converters).
-   * A README.md with any usage restrictions.
+   * The model in as many glTF variations as possible (using the same directory structure as the others ([example](2.0/Box))).  Tools for converting to glTF are [here](https://github.com/KhronosGroup/glTF#converters).
```
emackey: Instead of "possible" can we say "reasonable" here?  Large models omit variations to keep the size of the repo down.

README.md
javagl: Sure, thanks, I overlooked that. Fixed now.

README.md
javagl: I fixed it, but it is a bit subtle and subjective, of course: There are small models that are only available as `glTF` as well. But in any case, that's the baseline, and in doubt, the available tools usually make it easy to create the embedded/binary variants of it.

PR #201, opened 2018-11-07
Adding WaterBottle example that uses EXT_lights_image_based

MiiBond:
Here's an example of the WaterBottle asset using the EXT_lights_image_based extension


bghgary:
@MiiBond Do you have the source assets that can be contributed as well? I think that will be useful for this.

MiiBond:
You mean the HDR environment? It's actually the default in Dimension. Let me see if I can get the source...

bghgary:
You also need to add a README with the license info and a screenshot.

MiiBond:
If I'm going to add a README, should I create a new root folder called `sampleImages` for this?

bghgary:
Sorry, I didn't realize this is adding to the existing water bottle. Maybe it should be a different sample?

MiiBond:
It could be. But where would I put the HDR image?

bghgary:
> But where would I put the HDR image?

The source for models are typically here: https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/sourceModels

We have an asset that @PatrickRyanMS created ([see comment here](https://github.com/KhronosGroup/glTF/pull/1377#issuecomment-407202750)) that is just a bunch of spheres with different roughness values, one row of metal and one row of dielectric. I wonder if that would be a better asset for the IBL?

image

emackey:
> You mean the HDR environment? It's actually the default in Dimension. Let me see if I can get the source...

What's the license on the default HDR env in Dimension?

MiiBond:
The default HDR in Dimension has the Adobe Enhanced License.
https://stock.adobe.com/license-terms?prev_url=detail&comparison-full#enhanced-license-terms
It should be fine to distribute as long as anyone using it adheres to the license.

I'm still a little unclear about how I should include the original HDR in this repo if I'm just applying it to a model that already exists here. @bghgary are you suggesting that I create a new model folder but use the existing spheres model? Is it too confusing if I just add the HDR to the `sourceModels/MetalRoughSpheres` folder and edit the license to add separate info for the HDR image?

bghgary:
> are you suggesting that I create a new model folder but use the existing spheres model? Is it too confusing if I just add the HDR to the sourceModels/MetalRoughSpheres folder and edit the license to add separate info for the HDR image?

Yeah, maybe the metal rough spheres will be good. @PatrickRyanMS any thoughts on using that model?

emackey:
Actually I may have a better model for showing HDR reflections, a model of some chess pawns.  If you're willing to use that I can upload it soon.  One of the most important things it shows is how dark shiny surfaces interact with the environment.

PatrickRyanMS:
> are you suggesting that I create a new model folder but use the existing spheres model? Is it too confusing if I just add the HDR to the sourceModels/MetalRoughSpheres folder and edit the license to add separate info for the HDR image?
Yeah, maybe the metal rough spheres will be good. @PatrickRyanMS any thoughts on using that model?

@bghgary I agree that the spheres grid is good at showing the roughness falloff for your environment. It's purposely simple and repetitive because I use it to validate our DDS environment files and determine if the roughness falloff is correct when precomputing it. The model itself isn't anything special, but it is a standard practice in matching lighting conditions in post by dropping a chrome sphere into a scene so it's a mental model that is familiar.

emackey:
@PatrickRyanMS @bghgary I'd be fine with that model too, but I'd request an additional row of spheres showing a black dielectric (non-metallic) surface at different roughnesses.

The reasoning is that using an HDR environment looks mostly the same as using a non-HDR environment for models that are bright and shiny.  HDR comes into play when a dark surface is shiny, as it reflects the environment with a much lower exposure.  If you change the exposure of a non-HDR image, artifacts immediately appear in the overexposed areas, they turn into gray patches.  So, a row of black shiny spheres will reveal what's really happening in the HDR "brighter than white" spaces in your environment map.

emackey:
Agreed that intermediate values between metallic and non-metallic aren't needed.  Possibly I should remove them from the existing sample model, but that's a topic for another thread.

What your model doesn't show is a range of exposures.  Maybe we could add a third row, fully metallic, fully smooth, with grayscale base colors at `[0.04, 0.1, 0.2, 0.3, ... 1.0]` to show the environment at different exposures.

bghgary:
> Possibly I should remove them from the existing sample model, but that's a topic for another thread.

I'm not sure. I don't think we should have two sample models with metallic roughness spheres. Would it be okay to replace the existing with whatever we come up with here?

emackey:
> Would it be okay to replace the existing with whatever we come up with here?

I've had some great debugging sessions with the old model, but yeah, I guess a newer one would help steer people in a better direction (away from intentionally using fractional metallic values).

Would we still want to ensure visual consistency among engines for fractional metallic values?  Removing the sample model also removes it from the reference viewer, since they're using a git submodule.

emackey:
> I was thinking we replace the existing sample model with this one (i.e. the same folder and names)

Sure, but my point is, if we no longer have samples that show the full range of "partially" metallic values, then there won't be any models in the reference viewer to ensure consistency among the results of that.

That said, surely there are sample models that have these values along the fringes of where a metal meets a non-metal.  If it's OK for the reference images to be compared on just these fringes, instead of full spheres, then yes we can replace the old model with the new.

bghgary:
Oh, I see. Yeah, that might be an issue. Maybe use animations to hide them, but that might be weird.

emackey:
Here's my suggestion.  This model:  
[Pawns.zip](https://github.com/KhronosGroup/glTF-Sample-Models/files/2567406/Pawns.zip)

Paired with this HDRI:  https://hdrihaven.com/hdri/?c=indoor&h=pillars

@MiiBond Are you able to test out this pairing in Dimension, and share the result with us?

Here's a screenshot from BabylonJS:

pawnsandpillars

emackey:
> Dimension can't currently import glTF

Ah, so, on the roadmap for next version, right?  😄 

Is there an import path that preserves metallic/rough settings?  Also, how does one export, is it just File -> Export, or is there some deeper menu to visit?  One of my coworkers has a CC subscription, so I might be able to pester them to export something.

MiiBond:
Importing glTF is being discussed.

The only import path that preserves PBR settings is a custom OBJ import with embedded MDL. I think there might be a plugin for Substance to allow exporting of this but I'll have to check. Most of our workflows involve the user setting up their materials in Dimension.

Dimension 2.0 has a "Share 2D Scenes" option in the upper right corner. This will export a GLB to Creative Cloud Assets and you can view it in our viewer.
We're currently figuring out what functionality we're going to include for File->Export.

emackey:
So it seems the version of Dimension that I have access to doesn't produce correct files (they have Draco errors, which were corrected in a later version I believe).

Should we fall back to a set-of-spheres example?  I'm OK if this is a different/smaller set of spheres than the whole Metal/Rough test grid.  Or we could even use a model from the intro blog post.  Thoughts?

bghgary:
> Should we fall back to a set-of-spheres example?

How about the one from @PatrickRyanMS with an additional row of black spheres?

> I'm OK if this is a different/smaller set of spheres than the whole Metal/Rough test grid. 

I think ideally we have one metal/rough spheres model with a variant that has this extension in it.

I'm reviewing the reference viewer and having a model even without this extension to test is useful.

MiiBond:
Opened a new PR (#212) instead of this one.

PR #202, opened 2018-12-03
Added old camera model for testing.

UX3D-nopper:


emackey:
Should we rename?  It could be called "Antique Camera" or similar.

bghgary:
+1 for "Antique Camera" (though the folder should be without the space).

Issue #204, opened 2018-12-04
riggedSimple has noop animation channels


andreasplesch:
There are translation and scale targets which are animated but the values actually stay constant, and therefore do not need to be animated. This is probably the result of automated conversion. The example should be simplified as it aims to be simple.

lexaknyazev:
Duplicate of #86.

Issue #86, opened 2017-06-23
Skin animation models have animations and/or animation channels that do nothing


bghgary:
For example, RiggedSimple has two animations. The first one has linear interpolations that animate all the transforms to the same values for the entire timeline. The second one has a real curve for rotation, but translation and scale also have the same values for the entire timeline.

It would be good to update the models to at least remove the no-op animation from the list.

lexaknyazev:
Related: https://github.com/KhronosGroup/glTF/issues/982#issuecomment-304914349.

pjcozzi:
We could do this by running through [gltf-pipeline](https://github.com/AnalyticalGraphicsInc/gltf-pipeline) as soon as it has glTF 2.0 support.

emackey:
Some skinned models were regenerated in #245.  Safe to close this?

lexaknyazev:
As of now, `RiggedSimple` has one animation with three channels (TRS). Translation and scale do nothing.

lexaknyazev:
Long-standing issue in COLLADA2GLTF: https://github.com/KhronosGroup/COLLADA2GLTF/issues/225

DrX3D:
Untouched for almost 4 years. Can this be closed, or does it need to stay open because of the previous comment indicating an issue with COLLADA2GLTF? @lexaknyazev

donmccurdy:
Note that channels with a constant value are not necessarily unwanted. When the application is playing back multiple animations, or transitioning from one animation to another, that track provides an important distinction between:

- (A) this animation does not care where joint X is
- (B) this animation expects joint X to be in a specific location (which might or might not be the joint's rest pose)

In other cases, a constant track may be relevant to determining the start time or duration of an animation.

Probably neither of those situations was actually intended here, and it's just a COLLADA2GLTF bug, but I just want to point out that this wouldn't be something we should rule out from the repository.

***

I would be happy to make any of the following changes to the model, if desired:

- (A) remove constant channels
- (B) resample constant channels down from 50 to 2 keyframes
- (C) resample constant channels down from 50 to 1 keyframe

Currently glTF Transform does (B) by default (with `resample` command), and could be used in a scripting environment to do (A) or (C). gltfpack will do (A) by default, and has a CLI flag to do (C) instead.

javagl:
This has some connection to https://github.com/KhronosGroup/glTF-Validator/issues/200

emackey:
[...]

PR #205, opened 2018-12-05
RiggedSimple: remove no-op channels

andreasplesch:


bghgary:
I'm don't think mucking with the JSON directly is a good idea. Now there will be part of the binary that isn't referenced and the [source model](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/sourceModels/RiggedSimple) is out of sync. I think we should either fix collada2gltf to not output these or use some automated tool like gltf-pipeline to clean it up.

andreasplesch:
I think the source model is rather irrelevant. The collada data looks like it is actually not the original source but itself generated by an inefficient exporter itself. A main, perhaps the most important function of the sample models is that they can be inspected and learned from, with the knowledge that they are validated, and work with many glTF viewers and systems. It would be great if gltf-pipeline could be updated to recognize no-op animation channels or no-op/default translation, rotation, scale data.

bghgary:
The source model is how some of these models are generated by running it through collada2gltf. If we make an update to collada2gltf or the source model and there are manual changes to the resulting gltf, we won't be able to run an automated script to generate these models anymore.

emackey:
Fixes #86, Fixes #204 

@bghgary I'm on the fence about the need for future re-conversions of these models.  We do have a number of hand-made models, for example showing an individual triangle or box.  For more complex models, we now have at least one working glTF importer.  Do we really need to keep the Collada file around, and ensure that the provided sample can be automatically regenerated from it?

emackey:
> Now there will be part of the binary that isn't referenced

That would be a problem, yes.

bghgary:
> Do we really need to keep the Collada file around, and ensure that the provided sample can be automatically regenerated from it?

I don't have a strong preference, but if they are going to be out of sync, then we should remove the source.

bghgary:
Or state in the source that manual modifications were made.

andreasplesch:
I had added a step to the generator key value indicating manual modification. 

The .dae says it is exported from blender. It animates the full trafo matrix. Collada2gltf presumably could check during decomposition into  TRS if there is no change. Perhaps worth filing an issue with it.

If the .blend is available it may be easy to check how the blender exporter deals with it.

emackey:
Thanks @andreasplesch.  This model was rebuilt in #245.

In general, I think we should prefer fixing the conversion/export toolchain as opposed to manually fixing samples with hand-edits (except for just those few samples that were created by hand, which this one is not).

andreasplesch:
ok, this should rather be fixed here: https://github.com/KhronosGroup/COLLADA2GLTF/issues/225.
#86 tracks that also.

PR #206, opened 2019-01-06
add Interpolation Test

Reon90:
Spec: [animation sampler.interpolation](https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#animation-samplerinterpolation)


lexaknyazev:
@Reon90 
Please fix the validation error: 
```
/images/0: Recognized image format 'image/jpeg' does not match declared image format 'image/png'.
```
A proper action here is to save that image as an actual PNG file. Also `image.name` and `image.mimeType` could be safely omitted.

cx20:
I have tested this model with [gltf-test](https://github.com/cx20/gltf-test#simple-models-for-testing-individual-features).
The current status is as follows.
image

06wj:
@cx20 
As discussed in https://github.com/cx20/gltf-test/pull/48#issuecomment-446993361 , the animation of this model is not in the recommended format, so many engines only one box has animation.

emackey:
@06wj In this case I think it's helpful to be able to turn the animations on and off individually.  If they were all a single animation, there would be no way to reduce the amount of noise.

I suggested a new `allAnimations` flag in https://github.com/cx20/gltf-test/pull/57

bghgary:
> If they were all a single animation, there would be no way to reduce the amount of noise.

FWIW, there are asset generator tests that test individual animation types.
https://github.com/KhronosGroup/glTF-Asset-Generator/tree/master/Output/Animation_Node

PR #207, opened 2019-01-08
Add extension test for KHR_materials_unlit.

emackey:
New test model for KHR_materials_unlit.


2.0/UnlitTest/glTF/UnlitTest.gltf
```diff
@@ -0,0 +1,155 @@
...
+            "byteOffset" : 2304
+        }
+    ],
+    "buffers" : [
+        {
+            "byteLength" : 2568,
+            "uri" : "UnlitTest.bin"
+        }
+    ],
+    "extensions" : {},
```
bghgary:
```suggestion
```

2.0/UnlitTest/glTF/UnlitTest.gltf
```diff
@@ -0,0 +1,155 @@
...
+            "nodes" : [
+                0,
+                1
+            ]
+        }
+    ],
+    "extensionsUsed": [
+        "KHR_materials_unlit"
+    ],
+    "extensionsRequired": [
```
bghgary: This is true for the specGloss sample also. Is adding extensionsRequired a good thing for the sample models? I didn't do this for TextureTransformTest. At the least, I think the README should mention that the assets have the extensionsRequired set and that if the implementation doesn't support this extension, it is expected to fail to load. Thoughts?

2.0/UnlitTest/glTF/UnlitTest.gltf
emackey: This actually comes straight out of the glTF-Blender-IO exporter.  Safe to remove it of course, and I should pull it from the `glb` as well.

2.0/UnlitTest/glTF/UnlitTest.gltf
donmccurdy: @emackey Was this added by glTF-Blender-IO? I could see the argument for requiring the extension in a test file for that extension, but I definitely didn't meant to make it the default from the Blender exporter..

emackey: Yeah, the case was clear in the specGloss test, where I intentionally removed the metal/rough fallback from the spec/gloss side of the test, to prevent comparisons of metal/rough against itself.  So that model requires the specGloss model to load correctly.

This one is more subjective:  Certainly something will load and display if unlit is not required, but, the clear intention of this model is to display the unlit material itself, which can't be done without the support of the extension.

I'll add a note to the README.

2.0/UnlitTest/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+![screenshot](screenshot/screenshot_large.jpg)
+
+## Description
+
+This model tests the [`KHR_materials_unlit`](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_materials_unlit) extension.  There are two objects, one orange-ish (`#FF7F00` in sRGB space) and one blue-ish (`#007FFF` in sRGB space).  The visible surface of these two objects should appear as a uniform color, so much so that the 3D effect is intentionally lost.
+
+## Problem: Lighting applied to un-lit model
+
+![screenshot](screenshot/unlit_test_fail.jpg)
```
donmccurdy: minor: screenshots should say `KHR_materials_unlit` (plural)

bghgary: This image includes shadows. It might get confused with the "lighting effects are visible" statement. If you are planning on updating the screenshot, it might be better to do it without the shadows. If you want, you can easily turn off the unlit extension in the BabylonJS sandbox via the inspector.

image

emackey: Cool.  Off-topic, can this change be applied in VSCode?  At the bottom it says I need to reload the model, but in VSCode that reloads the entire webview, including resetting the inspector.

emackey: Also, on the Babylon Sandbox website, when I turn this off, then it (correctly) gives me the "extension not supported" error message 😄

bghgary: > can this change be applied in VSCode?

I was just thinking about that. We need a button to reload the asset for this to work.

bghgary: > Also, on the Babylon Sandbox website, when I turn this off, then it (correctly) gives me the "extension not supported" error message

Yeah, you will have to temporarily remove the `extensionsRequired` part.

emackey: Thanks!

2.0/UnlitTest/glTF/UnlitTest.gltf
emackey: README updated.

2.0/UnlitTest/README.md
emackey: Actually I think we should keep the shadows.  I got rid of the red text above, and the README below already talks about the shadow being allowed.

bghgary: Oh, I missed that. Sounds fine then.

cx20:
I have tested this model with [gltf-test](https://github.com/cx20/gltf-test#extension-test-models).
The current status is as follows.
image

PR #208, opened 2019-01-08
Rename the InterpolationTest animation names.

emackey:
This makes the names a little more meaningful.

Follows #206.

/cc @Reon90


emackey:
Travis CI had some problem but it doesn't look related to the simple change here.  Merging.

PR #210, opened 2019-01-09
 KHR_lights_punctual model

Reon90:
[Spec](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_lights_punctual)


emackey:
Thanks @Reon90.

This seems a little short on details.  The README should probably answer questions such as:  What is it trying to show?  What constitutes successful rendering vs not?  How can the viewer see if the effect of each light source ended up in the correct location or orientation?

cx20:
I tried displaying Lights.gltf in several libraries.
However, the screen shot of the sample seems to be different from the display result.

this model's screen shot:
image

[Three.js + Lights.gltf result:](https://gltf-viewer.donmccurdy.com/#model=https://rawcdn.githack.com/Reon90/glTF-Sample-Models/c5487f7a5ea616330597728d2bcdafc3b0101cb4/2.0/Lights/glTF/Lights.gltf)
image

[Babylon.js + Lights.gltf result:](https://sandbox.babylonjs.com/index.html?assetUrl=https://rawcdn.githack.com/Reon90/glTF-Sample-Models/c5487f7a5ea616330597728d2bcdafc3b0101cb4/2.0/Lights/glTF/Lights.gltf)
image

I do not know which is the correct result.

Reon90:
@cx20 but I made screen shot in blender 2.80, can't we trust it?

Reon90:
@cx20 I made point light closer.
screenshot 2019-01-12 at 03 34 07

cx20:
@Reon90 I could not find a model in `Lights.blend`.
image

Is the file correct? (Because I am not familiar with Blender, it may be misunderstanding.)

Reon90:
@cx20 I think this model can open only in blender 2.80, but your version is 2.79.

cx20:
@Reon90 Thanks. I was able to open that model with Blender 2.80 Beta. 
image

donmccurdy:
Unfortunately I don't think we can use Blender for the canonical screenshot here. There are open questions about the physical basis of their analytical lights' intensity (https://github.com/KhronosGroup/glTF-Blender-IO/issues/24).

I'll fix the issue in the three.js rendering shortly.

donmccurdy:
Updated three.js screenshot looks consistent with other images:

(although I forgot to disable IBL when taking these)

screen shot 2019-01-11 at 5 37 53 pm
screen shot 2019-01-11 at 5 38 04 pm

cx20:
@donmccurdy In the updated model the position of the `point light` got closer to the cube.
Can you check if the red `point light` is displayed in three.js?

06wj:
@cx20  When the range is not defined, Hilo's light is not attenuated, so the light will be very bright. I'll fixed it.

06wj:
@cx20 Fixed it.
[Hilo3d + Lights.gltf result:](https://hiloteam.github.io/Hilo3d/examples/glTFViewer/index.html?url=https://rawcdn.githack.com/cx20/gltf-test/2d27f4bd85dbd0cb2670c0d57d4d9c3e8baa0978/tutorialModels/Lights/glTF/Lights.gltf)

Reon90:
What about this pull request? Is there any concerns now?

emackey:
I have mild concerns that it's not easy for a casual observer to tell from a glance what the correct rendering of this is supposed to be.  But it seems like several bugs have already been found & fixed here, so it's probably worth merging.

cx20:
I think it would be nice to have a diagram explaining the position and direction of the light to help understand this model.
The following is the result of displaying the current model using light helper of three.js.
image

prideout:
This PR has been around for a while, any hope of getting it merged?  Would be nice to have a test for punctual lights.

cx20:
@prideout BTW, I am looking for a `Filament` + `glTF` import sample. If it exists, Could you tell me if there is one? I would like to add the sample to [gltf-test](https://github.com/cx20/gltf-test).

donmccurdy:
Along the lines of @emackey's comment https://github.com/KhronosGroup/glTF-Sample-Models/pull/210#issuecomment-452849188, I'm not sure this is a clear enough test of punctual light support. I'd prefer if we could create something that is more self-explanatory and unambiguous.

An example from https://github.com/KhronosGroup/glTF/pull/1223 (which probably needs to be recreated at this point) would be:

45334700-bdaba300-b530-11e8-9e7d-fd1dda53035d

emadurandal:
Hi, I think that we need common visual criteria. Why don't you choose a reliable physically-based offline renderer for getting a "ground-truth quality" rendered image?

I'm concerned about Blender's use of Watts (energy consumption? outgoing light energy?) for light strength.
If we can convert this scene data to Maya or something like that, We'll be able to render images in other proven offline renderers.

cx20:
@Reon90 Please let me know if you remember.

The point light for this Blender model is `500w`, but the exported glTF model has an intensity of `5`.
Are the intensity values in the output glTF file the result of manual adjustment?
When I tried it out with the latest version of the Blender glTF Exporter, the intensity also came out at `500`.

image

If the intensity values in the glTF file are more correct, we would be better off using the glTF Viewer results for the screenshots as well.
Or as @emadurandal points out, we might as well take a screenshot using a more physically correct renderer.

Reon90:
@cx20 I believe that it was originally 5 because according to spec the units are not a Watt:
```
The units that this is defined in depend on the type of light. point and spot lights use luminous intensity in candela (lm/sr) while directional lights use illuminance in lux (lm/m2)
```

cx20:
@Reon90  Thank you for your reply. I tried opening Lights.blend again with an older version of Blender.
Apparently, the light unit of Blender at that time was not watt. The mystery has been solved.

image

donmccurdy:
Yes, I think we were not sure of the units in Blender when KHR_lights_punctual support was written, but it's now clearly in watts. The bug to fix this (in Blender) is https://github.com/KhronosGroup/glTF-Blender-IO/issues/564.

cx20:
@donmccurdy Thank you for the information. I understand why you said you can't adopt the Blender screenshot.

cx20:
I tried again to view this model in some glTF Viewer.

|three.js|Babylon.js|Hilo3d|
|:---------:|:---------:|:---------:|
|image|image|image|
|Gestaltor|glTF Sample Viewer||
|image|image||

All the viewers seem to be generally correct, but the [Khronos glTF Sample Viewer](https://github.com/KhronosGroup/glTF-Sample-Viewer) seems to have too much light.

zeux:
It would be great to merge this (unsure what's preventing this, other than a conflict that needs resolution?), as right now there are no other examples of this extension in the repository.

emackey:
@zeux My problem with this model is there's no easy way to tell if a viewer got it right or not.  Lately we've been working to raise the bar for sample models, and I think this one is in need of a refresh, if not a redesign.  It should clearly label the different light sources, and have some means of indicating the intended effects.  It should also test to make sure that any angles or rotations that can be specified on lights have been implemented with the correct rotational direction and order of transformations.

donmccurdy:
This can be a stretch goal (I'd certainly accept a sample without it) but aspirationally it would also be great to have a sample that confirms whether `occlusionTexture` occludes punctual lights (it shouldn't).

zeux:
@emackey Mhm, I guess this makes sense although I wonder if presence of a test that exercises this functionality is better than absence of a test even if this is imperfect :) Unfortunately glTF-Asset-Generator doesn't contain a sample for this extension either which means that tools that want to make sure this extension is supported need to get the scene from somewhere else.

cx20:
I also think it would be better to have some kind of model to test this extension. A cooler model would be nice, but if it takes time to prepare, I think this model would be useful for minimal development.
However, the Blender file included in this PR is a bit out of date, so we might want to update to the latest version and re-export it.

prideout:
I agree with @cx20, it's important to at least have a simple smoke test for punctual lights.

DrX3D:
@Reon90 : Thre are conflicts in the branch that is used by this PR. Please resolve those conflicts and address any open concerns so that this PR can be merged.

javagl:
@DrX3D This PR has been open for >4 years. It is very likely that the person who opened is has moved on. Just in case: The conflict is only in the README (which is/was always an issue: Who is responsible for updating the tables? This will be sorted out in the new infrastructure). So the conflicting file could be ignored or resolved by just brushing the original README over the one that is changed here.

javagl:
My last comment was only about the fact that the merge conflict would become obsolete if this was just moved to the `glTF-Sample-Assets` repo.

Beyond that, I cannot say whether this model is "suitable" as a sample model. I think there _should_ be _at least(!)_ one sample model for each extension. But in [a recent comment](https://github.com/KhronosGroup/glTF-Sample-Models/pull/210#issuecomment-741508522) @emackey raised some concerns about the model itself, so it might not be the best one to use here...

donmccurdy:
That being the case, I think we should probably not merge this PR. Instead we should provide a model that either clearly illustrates correct vs. incorrect implementation, or showcases the artistic value of the extension.

Support for KHR_lights_punctual is much better in Blender to day than it was when this PR was opened, so that shouldn't be as much work as it might have been previously. Getting consistent punctual lighting between authoring and rendering tools is not trivial, for reasons related to image formation concepts like exposure and tone mapping — we should not add a sample at all without addressing those issues, at least in the description.

emackey:
On the artistic side of `KHR_lights_punctual`, we now have:

https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/LightsPunctualLamp

PR #216, opened 2019-03-01
Fix Linear Scale test in InterpolationTest.

prideout:
The animation whose name is `Linear Scale` is mistakenly using `CUBICSPLINE`.


donmccurdy:
The sampler layout will be different, if it is actually using CUBICSPLINE. Maybe we need to regenerate this from Blender.

prideout:
Ah that makes Don.  That also explains why the accessor has a wrong count of 3x.

emackey:
This is fixed now in #230.

PR #217, opened 2019-03-01
Fix LinearScale input accessor count.

prideout:
Validation and visual testing reveals that this animation expects 5 elements but the accessor contains 15 elements.


donmccurdy:
@Reon90 do you have the source `.blend` file for this? Seems like we may want to file a bug against the Blender exporter, unless the animation was manually edited after export or something.

emackey:
The count was too large because this animation was using CubicSpline even though its name was Linear.  Fixed in #230 (and the accessor being edited here is now unused, for reasons given in that fix).

Issue #219, opened 2019-03-13
pbrSpecularGlossiness BoxInterleaved sample does not contain any extension.


vpenades:
The model located [here](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/BoxInterleaved/glTF-pbrSpecularGlossiness) is defined as requiring pbrSpecularGlossiness.

But the actual material does not contain any extension material, it just uses the standard metallic material.

bghgary:
See related #151. I've just opened a PR to remove the bad models.

Issue #151, opened 2018-03-08
Models using spec/gloss materials are too glossy.


donmccurdy:
I feel like this has been discussed somewhere but I can't recall where, so filing here.

Models using spec/gloss appear too glossy:

| default | spec/gloss |
|--|--|
| <img width="407" alt="screen shot 2018-03-08 at 10 54 37 am" src="https://user-images.githubusercontent.com/1848368/37170510-dbcac6da-22bf-11e8-9aa1-229765bef02c.png"> | <img width="457" alt="screen shot 2018-03-08 at 10 54 32 am" src="https://user-images.githubusercontent.com/1848368/37170521-e00ab0e8-22bf-11e8-957c-b9df8ac22852.png"> |

Problem appears in both threejs and babylonjs. Most likely related to conversion in COLLADA2GLTF?

bghgary:
Yeah, https://github.com/KhronosGroup/glTF-Sample-Models/pull/76. There are a bunch of issues with the spec/gloss models.

donmccurdy:
Ok thanks, filed https://github.com/KhronosGroup/COLLADA2GLTF/issues/156 but will leave this issue open for those using these models to test.

bghgary:
Should we remove these models from the repo like we did for other models that are not correct?

donmccurdy:
I don't have bandwidth to look into COLLADA2GLTF right now unfortunately, so unless someone else is planning to, yeah probably better to remove these models.

emackey:
This can be closed now, right?

bghgary:
I say so. We can always reopen if we need to.

PR #220, opened 2019-03-13
Remove specular/glossiness models generated by COLLADA2GLTF

bghgary:
See https://github.com/KhronosGroup/glTF-Sample-Models/issues/151#issuecomment-372407413 and more recently #219.


emackey:
I'll update the model index after merging this.

Issue #222, opened 2019-05-07
AntiqueCamera uses different textures for glTF and GLB variants


lexaknyazev:
All six textures stored in `AntiqueCamera.glb` are 8-bit, while glTF variant has four 16-bit textures.

/cc @UX3D-nopper

lexaknyazev:
Apparently, @emackey has opened a #223 to fix this.

emackey:
Ah, didn't realize someone else had already been assigned.  Hopefully @UX3D-soares had not started work on this yet.

UX3D-nopper:
Actually, if you already solved it, please commit.

emackey:
Yes, I have a PR open with a fix, it's just waiting on the spec issue resolution.

emackey:
The spec issue is still open, but merging #223 fixes the sample model by avoiding 16-bit PNGs.

PR #223, opened 2019-05-07
Migrate AntiqueCamera textures to 8-bit PNG

emackey:
For https://github.com/KhronosGroup/glTF/issues/1561

The previous `glTF` version used 16-bit PNGs.  The previous `glTF-Binary` version did use 8-bit PNGs, but the embedded 8-bit PNGs appeared to be in the wrong colorspace.  For example, the two normal maps were a deep navy blue, much darker than a typical glTF normal map appears when viewed as an image.  The base color was likewise affected, and the resulting 3D model appeared too dark when compared with its own sibling.

This PR fixes the `glTF` folder to contain 8-bit integer PNGs, and replaces the .glb with a new one converted from the corrected glTF.  Both models now look the same and appear to be correct.


emackey:
Fixes #222.

PR #233, opened 2019-09-16
Update FlightHelmet sample to split hose into its own mesh

bghgary:
Update FlightHelmet sample to split hose into its own mesh with its own double-sided material.
Update source to be original Maya project.

This is a follow-up to #232.

There are a couple of issues with the glTF that still needs to be fixed before merging:
- [x] https://github.com/BabylonJS/Exporters/issues/626
- [x] https://github.com/BabylonJS/Exporters/issues/627


lexaknyazev:
@bghgary 
As both linked Maya exporter's bugs have been fixed, can we proceed with this PR? The updated model has a few validation issues: one empty node (from omitted light source) and three unused textures.

bghgary:
It looks like one of the bugs was closed but it wasn't actually fixed.

bghgary:
Also updated the scale so that the model matches real-world scale.

PR #236, opened 2019-10-03
fix draco accessor count and min/max values

### LINKED ISSUE #234: Some draco compressed models have invalid accessor count
I recently ran into the issue where the number of points coming back from `dracoGeometry.num_points()` did not match the value specified in `accessor.count`.

It seems that this is considered in the [spec](https://github.com/KhronosGroup/glTF/blob/1aae8e9084a5966c61f810dd3a3186fe084b642b/extensions/2.0/Khronos/KHR_draco_mesh_compression/README.md#implementation-note)
but some of the samples models are seemingly incorrect.

I see there is a related [issue](https://github.com/KhronosGroup/glTF-Sample-Models/issues/147) but it seems to be a bit more widespread than that one.

I discovered that the following models (glTF-Draco variants) appear to be incorrect:

- 2CylinderEngine
- BrainStem
- Buggy
- CesiumMan
- CesiumMilkTruck
- GearboxAssy
- Lantern
- Monster
- ReciprocatingSaw
- RiggedFigure
- VC

The issue I ran into was that my initial array buffers were not large enough to accept all of the values from the draco decoder. This results in very strange rendering with missing normals or positions.

My code (JS/WebGL) initializes all array buffers (and typed arrays) based on the accessor and it's not really feasible to resize after the decompression.

I have created a [gist](https://gist.github.com/shannon/5ecdb58d815bbe028348033dd0cbd6c4) that contains a breakdown of all the models and accessors where the counts don't match. As you can see, in some cases the descrepiency is rather large, in other cases the number of points returned from draco is less than the accessor count. In that case the result usually rendered ok but this means that I have much more memory allocated for the buffers than I need and wasted cycles on the draw.

Based on the discussion from #147 it seems that the exporters needed to be updated. That was quite a while ago so I don't know if it's already been done and these models just need to be updated.

shannon:
Fixes #234 

I took a stab at patching the draco model accessor counts with a small script. I patched the counts and the min, max values based on if the following condition was true:

```js
Math.abs(accessor.min[i] - dracoMin[i]) > Number.EPSILON
```

My hope was that this would remove any floating point error differences but it looks like the values returned from draco did appear to be different enough to warrant an update in some cases.

I suspect you may want to correct this from the model exports but these patches solved all of the issues I was having so I figured I would share.


shannon:
This is no longer needed with #259 merged in

PR #238, opened 2019-11-04
Add MetalRoughSpheres glTF 2.0 version without textures (metallicFactor+roughnessFactor)

gkv311:
New model does not come from "sourceModels", but has been generated via Draw Harness test script and exported into glTF format having presentation close to original sample.

I've found it frustrating that such useful sample model comes only with "textures mode".


emackey:
Thanks!  Please move this to its own `MetalRoughSpheres_NoTextures` folder, with its own `README` and license information and screenshot.  That way it can follow the naming conventions here, and be discoverable from the `model-index.json` file.

gkv311:
I have moved model into dedicated folder (and improved it a little bit - with better assembly structure).
Does it look better now? Can I do something more?

gkv311:
> It is tiny, only about 6 millimeters across.

You are right - didn't think about physical sizes and more focused on simpler math in Tcl script for this sample, as it is not supposed to be loaded with any other objects. Making it larger like 6 meters in size would be a little bit better, but don't want to spoil git history for unimportant changes.

Issue #240, opened 2019-12-04
According to the specs, TriangleWithoutIndices should not render anything


Raveler:
The specs say this about a gltf-file without a scene-field specified:

> When scene is undefined, runtime is not required to render anything at load time.
> Implementation Note: This allows applications to use glTF assets as libraries of individual entities such as materials or meshes.

Since TriangleWithoutIndices does not specify a scene, the default rendered view should be empty, right?

emackey:
I read that "not required to render" as being also not prevented from rendering.  It would be permissible to show a blank screen, but it would be equally permissible to treat the file's first scene as the default scene automatically.

The scene list is an area of the spec that could use some clarification.  For example: https://github.com/KhronosGroup/glTF/issues/1542 and https://github.com/KhronosGroup/glTF/issues/1547

For the sample repo, perhaps we should just fix the sample model to name the correct scene?

javagl:
This is related to (or basically a duplicate of) https://github.com/KhronosGroup/glTF-Sample-Models/issues/166 - OK, that was my responsibility. I'll re-insert this into my TODO list with higher priority (and leave this issue open until it's *really* done). Sorry about that.

Raveler:
Alright, thanks! I know it's total nitpicking but those simple models are all used to pick up very detailed flaws in the implementation, so people implementing glTF might think this is such a corner case being tested :).

Issue #166, opened 2018-04-28
The sample models should all have a scene property


javagl:
As reported in https://github.com/KhronosGroup/glTF/issues/1326 , some of the models (particularly the hand-crafted `Simple...` models) may not have a `scene` property. According to the 2.0 spec, this means that nothing has to be rendered at all. (This was not yet entirely clear in 1.0). 

The corresponding samples should be reviewed and updated accordingly.

emackey:
We should probably also have a sample that has multiple scenes, to test what that does in various engines.

javagl:
Agreed. There had been some trouble with some glTF 1.0 viewers when there was no default `scene` (crashes and bugs). But I think that even more viewers (1.0 and 2.0) would bail out when there are *multiple* scenes. I'll try to create such a model as well, that could show some interesting effects ;-)

javagl:
Closed via https://github.com/KhronosGroup/glTF-Sample-Models/pull/242 .

The point of having *multiple* scenes in one asset is currently discussed in https://github.com/KhronosGroup/glTF/issues/1542 (shortly before I was pointed to that issue, I created a simple `MultipleScenes` example that currently only is in a branch at https://github.com/javagl/glTF-Sample-Models/tree/MultipleScenes )

PR #242, opened 2019-12-11
Added "scene" property to simple test models

javagl:
According to the specification, the runtime is not required
to render anything when the scene property is undefined.

Fixes https://github.com/KhronosGroup/glTF-Sample-Models/issues/240

### DIFF
--- a/2.0/AnimatedTriangle/glTF-Embedded/AnimatedTriangle.gltf
+++ b/2.0/AnimatedTriangle/glTF-Embedded/AnimatedTriangle.gltf
@@ -1,4 +1,5 @@
 {
+  "scene" : 0,
   "scenes" : [
     {
       "nodes" : [ 0 ]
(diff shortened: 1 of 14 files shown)


emackey:
Also fixes #166 I believe.

javagl:
@emackey In some way, yes, but 166 also considered adding an example with *multiple* scenes. (I think there is **no** sample right now with multiple scenes, so I'm currently creating one - as usual for the "simple" test models, it's a trivial one, just to test which viewers choke on that (and in which way...))

emackey:
@javagl Actually I'm not sure the WG wants an official sample until we know what the plan is for multiple scenes.  There's an issue open for considering deprecating multiple scenes.

javagl:
@emackey Thanks for that hint. I also just had a look at the issue, but have not yet read through the discussion there - it *seems* like this is targeting glTF-next, and so one *might* consider adding the sample anyhow. But in doubt, for now, it resides in a branch at https://github.com/javagl/glTF-Sample-Models/tree/MultipleScenes , and 166 can probably be closed for now.

PR #247, opened 2020-01-17
Add quantized versions of a few models

### LINKED ISSUE #246: Sample of KHR_mesh_quantization needed
@zeux I don't suppose you have a public asset that you could contribute?

zeux:
Added AnimatedMorphCube, Avocado, CesiumMan and Lantern, processed using
gltfpack 0.13 to produce examples of KHR_mesh_quantization

Each model was generated using

gltfpack -i inputfile -o outputfile

and then reformatting the output file using a JSON prettifier since
gltfpack produces minified JSON by default.

Fixes #246


zeux:
I've included Avocado and Lantern as requested by @emackey, and also added CesiumMan (it's a skinned model) and AnimatedMorphCube (it's a model with position/normal/tangent morph targets) as well so that we have a full coverage for this extension in terms of elements that may be quantized.

All 4 models render and validate without errors with https://gltf-viewer.donmccurdy.com/. The CesiumMan model generates 2 warnings due to the skinned mesh being parented to a non-root node, but it doesn't look like this repository in general is warning free.

It looks like this can only be merged after the validator version used in this repository is bumped to the most recent one that supports KHR_mesh_quantization though.

lexaknyazev:
Original `CesiumMan` model has non-normalized vertex influences and will be replaced as per  https://github.com/KhronosGroup/glTF-Sample-Models/pull/245#issuecomment-574820729.

Maybe choose another skinned model for this PR to merge it sooner?

zeux:
Ah, I forgot about this. gltfpack renormalizes vertex influences - the model still looks fine upon visual inspection. I could choose a different model, but the model this PR adds passes validation (with a recent validator).

lexaknyazev:
Following the discussion from the linked comment, it seems that besides fixing weights, the original model will be facing different direction. This means that the quantized version will have to be regenerated to keep it aligned.

Up to @emackey what to merge first.

zeux:
Replaced CesiumMan with Monster for now - let me know if we'd rather merge CesiumMan or some other model instead.

lexaknyazev:
As listed in https://github.com/KhronosGroup/glTF-Sample-Models/pull/245#issue-361461416, all five skinned sample models from this repo have ignored transforms in their node graphs. Three of them (`CesiumMan`, `RiggedFigure`, and `Monster`) additionally have non-normalized weights which is considered a severe issue.

@emackey 
What do you think of priorities here?

emackey:
I think for now we should leave the skinned meshes off of this PR, at least until after #245 is settled and merged.

That said, if we can resolve https://github.com/KhronosGroup/glTF-Sample-Models/pull/245#issuecomment-575644292, perhaps that issue could come in before this one, in which case we could include skinned quantized models.

zeux:
Sounds good! Updated the PR to remove skinned models. We can always add them later.

emackey:
@zeux Thanks for this PR.  `master` now has a new glTF Validator that can handle mesh quantization, so if you merge or rebase here, this PR should get a green checkmark on it.

Also can you throw in that Duck model?  Thanks!

donmccurdy:
Note that all these models are simply generated with `gltfpack -i input.glb -o output.glb`... We can very easily add models incrementally.

zeux:
Added Duck to the PR. And yeah if we ever need more models I'm happy to convert them, this is just using `gltfpack` as Don noted, and running a JSON formatter after that since gltfpack is producing minified glTF files and I figured that the sample model repository should have files that are more readable.

lexaknyazev:
@emackey @donmccurdy 
OK to merge?

PR #249, opened 2020-01-29
Add the low-poly fox model

emackey:
Based on the one @scurest cleaned up in #150.


emackey:
I completely removed the old "Monster" model in #245, as it had serious issues beyond just conversion errors, such as a broken animation cycle, bad scaling, bad front direction, etc.  This fox appears to be a much better sample on many levels.

prideout:
Maybe the README for this sample should mention the flat-shading aspect of the model, and that an implementation is incorrect if draws the fox with smooth shading?

prideout:
Btw thanks for replacing the Monster model, this is much better :)

emackey:
Yes, this model lacks normal vectors, which is unintentionally controversial as I don't think we previously had any models here that were both lit and lacking for normals.

The spec states that clients are expected to compute flat-shaded normals in this case.  But this doesn't make sense for certain clients, as it adds not only the processing of the normals but also the need to split apart shared vertices.  There are team members on both Cesium and STK who argue that the lack of normals would be better interpreted as being unlit, rather than requiring all the processing overhead of splitting shared vertices and calculating all the flat normals.  But that's too large of a breaking change to retrofit into the official glTF 2.0 spec, so that issue is at an impasse for them.

In any case, the fox is here to demo multiple animation cycles, not lack of normals.  Probably I should add the normals to the fox, and make a separate model to test for lack of normals.

prideout:
Interesting!  Thanks for explaining the controversy.  For what it's worth, here on the Filament team we received a bug report ([link](https://github.com/google/filament/issues/2088)) due to the lack-of-normals in this model, so I can imagine other renderers running into the same thing.

Anyway, I think your plan to split this into two tests makes sense and thanks again for adding this model -- it's a nice test!

emackey:
Awesome!  When you get a chance, let me know what your team thinks of #251.  Thanks!

Issue #253, opened 2020-02-10
Animated cube example rotating opposite direction?


Leadwerks:
Both Windows 3D Viewer and my own code I just implemented show the cube rotating clockwise, but the GIF shows it going the other way.
https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/AnimatedCube

I suggest fixing or removing the image, if it is in fact incorrect, because developers are likely to use this model as their first test for animation.

bghgary:
The GIF does indeed look like it's rotating the wrong direction. @UX3D-nopper, I believe you added this asset? Can you confirm?

If you are testing your client implementation, I would suggest also using [glTF-Asset-Generator](https://github.com/KhronosGroup/glTF-Asset-Generator) for testing.

See [Animation_Node](https://github.com/KhronosGroup/glTF-Asset-Generator/tree/master/Output/Positive/Animation_Node) and [Animation_NodeMisc](https://github.com/KhronosGroup/glTF-Asset-Generator/tree/master/Output/Positive/Animation_NodeMisc) for node animations.

UX3D-nopper:
Yes, I added this asset during the specifaction of glTF 2.0. The animated glTF cube is rotating clockwise. The animated GIF is wrong.

Leadwerks:
Here's a quick fix:

![screenshot](https://user-images.githubusercontent.com/108503/74244121-9fc80c00-4ce9-11ea-8aea-eb29f91305e3.gif)

UX3D-nopper:
Suggest u do a pull request.

emackey:
Fixed in d7d0bd16684c1393a.

PR #254, opened 2020-02-15
Fixed image with rotation in wrong direction.

Leadwerks:
https://github.com/KhronosGroup/glTF-Sample-Models/issues/253


cx20:
The screenshot of `BoxAnimated` is included. Is it a mistake?
Looks like the same screenshot as `AnimatedCube`.

Leadwerks:
Ah, that is why the image was not refreshing. My mistake.

emackey:
Thanks @Leadwerks!  I cherry-picked your second commit with the correct gif filename.

PR #263, opened 2020-06-04
Added KHR_xmp metadata to the Damaged Helmet sample.

weegeekps:
Added KHR_xmp metadata for the Damaged Helmet sample to the glTF, glTF-Binary, and glTF-Embedded files.

KHR_xmp is the new extension outlined in [PR #1553 for glTF](https://github.com/KhronosGroup/glTF/pull/1553).

If you download these files you can see the KHR_xmp metadata using the Babylon Sandbox.

**Note:** This is using the `model3d` namespace which we haven't published the schemas yet for, so we may want to hold off merging until we have those published. The idea with opening this PR now is to provide a real world sample for people to look at.


bghgary:
I wonder if we should be adding this to the existing glTF or adding a new folder `glTF-xmp`? The `glTF` folder has so far only contained core spec assets.

weegeekps:
@bghgary I was wondering that as well, but the way @emackey was talking during our call two weeks ago it sounded like there is a desire to include the identifying and license metadata within the core files. I'm open to either approach.

emackey:
@bghgary That's a good point, and I didn't really consider the ramifications of adding a whole lot of non-pure-core assets to the main example folder there.

That said, we do expect core implementations to at least be able to safely ignore non-required extensions.  Also, most extension test models (clearcoat, texture transform, etc) do re-use the `glTF` folder with extensions applied.

I don't think we have a rule for when extensions go in the main glTF folder, vs when they use a separate folder such as `glTF-Draco` or spec/gloss.  My impression is that there are different testing strategies involved.  For example, with Draco we wanted to compare a non-Draco model to a Draco version of the same model.  Likewise with spec/gloss.  But with clearcoat, there's no use comparing to a non-clearcoat version of the same, so it gets its own test model and uses the main folder.  Likewise with tex transform.

In the case of XMP, the model isn't supposed to look any different, and isn't stored any differently in the file.  I don't think it's useful to compare an XMP model to a non-XMP version of the same.  So I don't think it makes sense to have a `glTF-XMP` flavored folder alongside a non-XMP folder.

Perhaps we could have one "XMP Test" model that really tries to push XMP through as many of its paces and permutations as could make sense together in a single file.

Is there a risk of polluting the sample models by adding non-required XMP to some of the core samples?  Certainly we expect non-XMP-supporting viewers to still read and render all of these models, regardless of the XMP data.

2.0/DamagedHelmet/glTF/DamagedHelmet.gltf
```diff
@@ -1,204 +1,242 @@
...
+          ],
+          "dc:relation": [
+            "https://www.khronos.org/",
+            "https://sketchfab.com/models/b81008d513954189a063ff901f7abfe4"
+          ],
+          "dc:rights": {
+            "en-us": "Battle Damaged Sci-fi Helmet - PBR by theblueturtle_(https://sketchfab.com/theblueturtle_), published under a Creative Commons Attribution-NonCommercial license."
+          },
+          "dc:title": {
+            "en-us": "Damaged Helment"
```
emackey, line 235: "Helment" has a typo.  Also, the full title is:  `Battle Damaged Sci-fi Helmet - PBR`

2.0/DamagedHelmet/glTF/DamagedHelmet.gltf
```diff
@@ -1,204 +1,242 @@
...
+        {
+          "dc:description": {
+            "en-us": "The original model was built on an early draft of glTF 2.0 that did not become final. This new model has been imported and re-exported from Blender to bring it into alignment with the final release glTF 2.0 specification."
+          },
+          "dc:format": "model/gltf-binary",
+          "dc:language": [
+            "en"
+          ],
+          "dc:relation": [
+            "https://www.khronos.org/",
```
emackey, line 228: What's the relation to Khronos here?  (Other than we've over-used this model for a long time...)  Should this instead be a link to `https://sketchfab.com/theblueturtle_`?

2.0/DamagedHelmet/glTF/DamagedHelmet.gltf
```diff
@@ -1,204 +1,242 @@
...
+      "@context": {
+        "dc": "http://purl.org/dc/elements/1.1/",
+        "model3d": "https://schema.khronos.org/"
+      },
+      "packets": [
+        {
+          "dc:description": {
+            "en-us": "The original model was built on an early draft of glTF 2.0 that did not become final. This new model has been imported and re-exported from Blender to bring it into alignment with the final release glTF 2.0 specification."
+          },
+          "dc:format": "model/gltf-binary",
```
emackey, line 223: I probably need to study the discussion on the XMP PR, but, is this format correct for the non-binary variant of this model?

And separately, is `format` something that converters will need to edit, when models are converted from *.gltf to *.glb and vice versa?  I'm a little concerned that GLB/gltf converters would need to edit the metadata when converting, that feels off.

bghgary:
I'm okay with this. I just want to make sure we discuss it. Something to consider also is that we are adding to the process of creating a model. If any of these models with metadata need to be updated, we have to remember to re-add the metadata. Maybe we should add the source metadata files to the source folder also?

emackey:
@weegeekps I'm going to close this for now. I'm sure we'll chat about establishing some samples for KHR_xmp_json_ld once the dust settles there.

PR #265, opened 2020-06-16
Add example model for KHR_materials_transmission

MiiBond:


cx20:
To be consistent with other samples, it is better to name the sample model file the same as the folder name is a good idea.

|Folder Name     |File Name|
|:---------------|:--------|

MiiBond:
> To be consistent with other samples, it is better to name the sample model file the same as the folder name is a good idea.

@cx20 Good point. Done.

cx20:
I tried adding this model to [gltf-test](https://github.com/cx20/gltf-test) as a trial.

image

So far, only [Babylon.js](https://cx20.github.io/gltf-test/examples/babylonjs/index.html?category=tutorialModels&model=TransmissionTest&scale=3.0&type=glTF) seems to support this extension.

BTW, it may be a matter of taste, but I prefer the model to be in the center of the Scene, as in the [ClearCoatTest model](https://cx20.github.io/gltf-test/examples/threejs/index.html?category=tutorialModels&model=ClearCoatTest&scale=0.3&type=glTF).

2.0/TransmissionTest/glTF/TransmissionTest.gltf
```diff
@@ -0,0 +1,1471 @@
...
+                    "dc:format": "model/gltf+json",
+                    "dc:title": "ASM_transparency_labels.dn",
+                    "xmp:CreatorTool": "Adobe Dimension"
+                }
+            ]
+        }
+    },
+    "extensionsUsed": [
+        "KHR_xmp",
+        "KHR_texture_transform",
```
emackey, line 1468: Echoing Alex's concern from https://github.com/KhronosGroup/glTF/pull/1698#discussion_r442783701, I don't think the official sample model for transmission should depend on `KHR_texture_transform`.  Can it be removed from this model?  XMP is fine to keep, as it's optional, but texture transform is required here for the correct labels.

In other models that require it, Adobe Dimension should probably list `KHR_texture_transform` under `extensionsRequired` given that the texturing looks broken in viewers that don't support it.  Viewers shouldn't attempt to load the model at all if they can't process the texture transform extension.

2.0/TransmissionTest/glTF/TransmissionTest.gltf
```diff
@@ -0,0 +1,1471 @@
...
+    "scene": 0,
+    "extensions": {
+        "KHR_xmp": {
+            "@context": {
+                "dc": "http://purl.org/dc/elements/1.1/"
+            },
+            "packets": [
+                {
+                    "dc:date": "2019-10-25T15:25:01.695-07:00",
+                    "dc:format": "model/gltf+json",
```
emackey, line 1459: Over in the XMP discussions, I think we talked about omitting `dc:format`.  Otherwise, all GLB <-> .gltf converters must be updated to search for `<namespace>:format` and update the format during conversion.

2.0/TransmissionTest/glTF/TransmissionTest.gltf
```diff
@@ -0,0 +1,1471 @@
...
+        }
+    ],
+    "samplers": [
+        {},
+        {},
+        {},
+        {},
+        {},
+        {},
+        {}
```
emackey, line 1384: Is there some way to prevent all these empty samplers?  At most a single default sampler could be shared here.

Apologies if I'm reviewing the Dimension exporter itself more so than this particular sample, we could certainly merge this without optimizing the samplers if needed.

2.0/TransmissionTest/glTF/TransmissionTest.gltf
MiiBond: Will do.

2.0/TransmissionTest/glTF/TransmissionTest.gltf
MiiBond: Yeah, @bghgary mentioned that to me as well and it's on my list of things to fix :)

2.0/TransmissionTest/glTF/TransmissionTest.gltf
MiiBond: I'll work on this. Might be a couple of days.

MiiBond: Okay, done.

emackey:
@lexaknyazev It looks like CI didn't run here.  Any ideas?

2.0/TransmissionTest/glTF/TransmissionTest.gltf
```diff
@@ -0,0 +1,1522 @@
...
+            "mimeType": "image/png"
+        },
+        {
+            "name": "texture927751",
+            "uri": "TransmissionTest_images/texture927751.jpg",
+            "mimeType": "image/jpeg"
+        },
+        {
+            "name": "texture927748",
+            "uri": "TransmissionTest_images/texture927748.jpg",
```
emackey, line 402: This image is flagged by the glTF Validator as not being used, and not being present on disk.

It should have failed CI with this, but it looks like CI isn't running.

emackey:
@lexaknyazev Correction, CI is running, just not showing up as green checkmarks or red X's on this PR.  It's not warning me not to merge the PR with an error in it.

https://travis-ci.org/github/KhronosGroup/glTF-Sample-Models/builds

2.0/TransmissionTest/glTF/TransmissionTest.gltf
MiiBond: Thank you. I was editing it by hand before submitting and wasn't careful enough.

emackey:
GitHub isn't showing it, but validation has passed: https://travis-ci.org/github/KhronosGroup/glTF-Sample-Models/builds/702172431

cx20:
I tried displaying this model in [Babylon.js Sandbox](https://sandbox.babylonjs.com/) again. However, if you apply the glTF's `built-in camera` on the latest version of the model, it doesn't seem to fit in the frame.

Previous model result:
image

Current model result:
image

I would suggest removing the built-in camera or adjusting it to fit in the frame.

abwood:
I love this model! After taking a close look, I've got a few questions regarding the sample.

I noticed that in cases where a transmission texture is used (e.g., RedTransTexture), we specify a transmission texture but transmission factor is not specified. Does it need to be? The spec says that transmissionTexture and factor are to be combined, but transmissionFactor defaults to 0, resulting in no transmission.

"GreenMask" material is using alphaMode="Blend". Should alphaMode be set to Mask? The texture is indeed a masking texture. There are possibly other materials that could be set to Mask rather than blend. The more of these spheres that render in the opaque pass, the better.

A final minor comment is that all meshes are named "Sphere". For debugging purposes it would be helpful if these had unique names.

MiiBond:
Thanks for catching that. I've updated `transmissionFactor` where appropriate.
The mask textures as basically black and white so, yeah, they could probably use "MASK". However, the point of including them was to test that a renderer is combining the transmission logic and alpha-blending transparency logic correctly. I could make these textures use more grey values to extenuate this.
I've renamed the spheres too.

donmccurdy:
In several renderers I'm seeing some issue that looks like imprecision in the alpha (UVs too close to a texture boundary?).

Screen Shot 2020-07-31 at 1 05 10 PM

The screenshot above is from three.js, which doesn't support the transmission extension yet, but the same thing appears in BabylonJS, which does.

emackey:
@donmccurdy I should double-check, but that error looks very, very familiar to me.  It's what happens when you have a double-sided sphere, with BLEND enabled and the z-buffer writes disabled, with no depth sorting.

I even have a hacky workaround for such models: Make two co-located single-sided spheres, as separate primitives.  Either the inside or the outside will be consistently "on top", which looks better than the bands even if the order comes out wrong.  But I don't know if such hacky workarounds belong in official sample models, probably they don't.

elalish:
Yes, in fact I implemented something very similar in <model-viewer> to
handle transparency better:
https://github.com/google/model-viewer/blob/ec2057513beb40a3c6ea9c2be0ea66d3065f4990/packages/model-viewer/src/three-components/gltf-instance/ModelViewerGLTFInstance.ts#L90,
which is based on a three.js example. My point here is I don't think this
adjustment needs to be part of the glTF; it can be applied in the
rendering engine as appropriate.

-Emmett

donmccurdy:
@emackey Nice guess, you're correct that the model is using `doubleSided: true` and `alphaMode: BLEND` in some materials. I would've expected this model to only require single-sided materials and a combination of OPAQUE and MASK alpha modes, though.

Screen Shot 2020-08-03 at 11 29 51 AM

^table does not include `transmissionTexture`, as `gltf-transform inspect` doesn't detect that yet.

abwood:
Yes, I can confirm this is an issue with BLEND mode, and the "easiest" way to fix this is to render doublesided transmission objects in two passes, backfaces then front faces.  I raised the BLEND vs MASK point earlier in this PR, but the choice of BLEND mode is intentional and likely helped us uncover a bug in another engine as a result anyway.  I think many of these spheres look better as doublesided anyway, as these patterns on the backside should be visible through the spheres.

MiiBond:
Yes, double-sided is very intensional as these are thin-shells where you can see the inside surface.
We could include a alpha map with more greyscale values so that the choice of `BLEND` setting is more clear but I don't feel strongly about this.

donmccurdy:
From the proposed spec:

> If alpha-as-coverage is not being used, the blend mode of the material should be set to "OPAQUE" even though it is transparent.

We don't typically expect transmissive or non-opaque materials to appear behind transmissive materials, do we? I understand that rendering doublesided transmission in two passes is a useful technique, but I worry that this model is testing many things not required in the spec:

1. OPAQUE materials visible through transmission
2. BLEND materials visible through transmission
3. other transmissive object visible through transmission
4. same transmissive object visible through transmission


(1) and (3) are fine, I'm wondering about (2) and (4). I'd rather not mix an alpha map with the sample, no. But should some of this be spelled out in the spec? Do we expect all clients to split doubleSided materials into two passes?

donmccurdy:
> However, the point of including [BLEND] was to test that a renderer is combining the transmission logic and alpha-blending transparency logic correctly. 

I might still prefer that this use MASK, rather than BLEND. It seems better for this example to demonstrate best practice than to exercise all the edge cases, and if we're deviating from best practice _in order to show an edge case_ we should be very clear about that in the README. But I think it would be easier to exercise those combinations in a separate example more like https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/AlphaBlendModeTest, later.

emackey:
@donmccurdy I'm not sure BLEND vs MASK makes much difference to this model, although I certainly agree we should have a separate test model to break out specific edge cases.  Off-topic, but not everyone on the Filament team is on board with the way we did MASK, and there are separate open issues on that.

Can this sample go in as-is?

donmccurdy:
> I'm not sure BLEND vs MASK makes much difference to this model....

That's almost the point — in a world where all engines support BLEND triangle sorting perfectly, you'd practically never need MASK, BLEND would always work fine. But implementation is not that simple, and so if you _could_ use either, you should choose MASK.

I can't see any reason to use BLEND here, except to make implementation more difficult. I don't think that is the right goal for our sample models, which should aim to show best practice and be very clear about the reason when they do not. We deal with three.js users (very regularly) having trouble with models using BLEND that could be using OPAQUE or MASK, and I don't want to encourage that as the default for exporters with the new transmission materials. 😕 

More generally: if alpha-as-coverage does not need semitransparency, the blend mode should be "MASK" and not "BLEND", as if the transmission property were not involved at all.

emackey:
@donmccurdy Good points here.

@MiiBond You did say in an earlier comment here that the use of BLEND is intentional, but BLEND is known to be problematic when multiple polygons overlap, such as on a double-sided sphere, even in the absence of transmission.  I think we need to test transmission with BLEND on geometry that doesn't overlap, which would mean a different test model than this.  In light of that, would it be OK to switch this model to use MASK?

MiiBond:
Hi. Sorry, I've been neglecting this PR.

Yeah, MASK is definitely more appropriate. I usually export with BLEND because it gives nicer edges and Dimension's web viewer does OIT.

cx20:
The following is the comparison result between the old and new models. The screenshot is a link to a sample. I can't determine if this is an expected result, but Babylon.js does show an improvement.

||Before (BLEND)|After (MASK)|
|:--:|:---:|:--:|
|Babylon.js|[image](https://rawcdn.githack.com/cx20/gltf-test/83486d578471406bd40edb675c39df2d03eff653/examples/babylonjs/index.html?category=tutorialModels&model=TransmissionTest&scale=5&type=glTF)|[image](https://rawcdn.githack.com/cx20/gltf-test/f2204726c00eb6895e3efc9ad82fce3eee49b950/examples/babylonjs/index.html?category=tutorialModels&model=TransmissionTest&scale=5&type=glTF)|
|Filament|[image](https://rawcdn.githack.com/cx20/gltf-test/fb1a0e3fcbfd7af9cee753dbff25ddf20e6f97aa/examples/filament/index.html?category=tutorialModels&model=TransmissionTest&scale=5&type=glTF)|[image](https://rawcdn.githack.com/cx20/gltf-test/f2204726c00eb6895e3efc9ad82fce3eee49b950/examples/filament/index.html?category=tutorialModels&model=TransmissionTest&scale=4&type=glTF)|

donmccurdy:
Thanks @MiiBond! Here's a summary of the materials in the latest draft:

Screen Shot 2020-09-14 at 2 48 21 PM

The only thing I might consider changing would be to replace the emissive "LabelMaterial" with an unlit material — it will behave better in path tracers or in rasterizers that use bloom effects. I'm not as concerned about that, so it's fine with me to merge this!

donmccurdy (approved):

MiiBond:
Re: unlit material. Are we okay including another extension in this sample model? I thought we were trying to avoid that.

emackey:
Yeah, I think it's cleaner if we leave unlit out, so we're only testing the one extension here.

donmccurdy:
Ok, works for me. 👍

PR #267, opened 2020-07-29
TextureTransformMultiTest

emackey:
This tests 2 UV sets (TEXCOORD_0 and TEXCOORD_1) in combination with KHR_texture_transform, across a variety of texture inputs.


emackey:
BTW, I'm very happy with the Blender exporter's output for this test model.  Many of my test models require careful hand-tweaking to reach some obscure edge cases, but, this particular model came straight out of stock Blender 2.83 with no hand-edits required.  The .blend file is included under `sourceModels`.

abwood:
As always, these models are well crafted and incredibly useful!  Out of curiosity, is there a case where the Unlit row would have failures that did not match the baseColor row?

emackey:
Thanks.  Yeah, the Unlit row might be redundant, as it reuses the baseColor input.  I don't think it hurts to leave it in though.

I might have to make another chart for the PBR Next extensions, once they're stabilized, but I'll make a separate PR for that when the time comes.

emackey:
@cx20 Can you test this one?  Thanks!

cx20:
@emackey Thank you for your new cool model. I experimentally added this model to https://cx20.github.io/gltf-test/.
Perhaps [Babylon.js](https://cx20.github.io/gltf-test/examples/babylonjs/index.html?category=tutorialModels&model=TextureTransformMultiTest&scale=1&type=glTF) and [Filament](https://cx20.github.io/gltf-test/examples/filament/index.html?category=tutorialModels&model=TextureTransformMultiTest&scale=1&type=glTF) are passing the test. However, the check mark on Clearcoat is a little dark, so it seems difficult to understand.

|Babylon.js|Filament|
|:------:|:------:|
|[image](https://cx20.github.io/gltf-test/examples/babylonjs/index.html?category=tutorialModels&model=TextureTransformMultiTest&scale=1&type=glTF)|[image]((https://cx20.github.io/gltf-test/examples/filament/index.html?category=tutorialModels&model=TextureTransformMultiTest&scale=1&type=glTF))]|

emackey:
@prideout I have a strange comment on the above screenshots.  Notice the "Normal" row of checkmarks, in Babylon the three of them look nearly identical.  But in most other engines, including Filament, the one in the last column (the non-texture-transform column), the reflections look different from the other two.

I believe this is because Babylon is using KHR_texture_transform to transform UV coordinates prior to calculating the full TBN matrix for the normal-tangent space.

In the case of Filament, the checkmark means that Filament is correctly transforming the actual UV coordinates used for normal mapping, but I think that the TBN matrix has maybe been generated from the original, un-transformed UVs.  So, the reflections are rotated 90 degrees in the "UV 0" and "UV 1" test columns.  This is just a guess, of course, but it's informed by my own efforts to implement support for this in a desktop engine.  I had to fix my TBN to get results like Babylon.

prideout:
Hi @emackey, you're correct -- Filament's glTF loader library generates a surface orientation based on the original un-transformed UVs.  Sounds like you found a bug!  Do you know if the desired behavior is specified somewhere?  I took a quick glance at KHR_texture_transform and don't see it in there.

emackey:
I'm sure it's not specified, anywhere.  I suppose some specification should be added to the KHR_texture_transform documentation.  Actually I already have a PR open against that doc for a different problem, but it's difficult to explain the orientation of the math in there, especially when FlipY gets involved.

I presume that if we updated the spec, transforming the UVs prior to calculating the missing tangents would make sense.  At least, that allows a standard glTF tangent-space map (red on the right, green from the top) to work correctly even when UVs are transformed.  But I wonder what happens when UVs are transformed and the model supplies its own tangent vectors?  The official sample viewer takes any supplied tangents verbatim, and doesn't attempt to transform them the way it does with calculated tangents.  Maybe I need to make a model to test that too?

emackey:
@cx20 I updated the Clearcoat boxes a little.  I'm not sure it's going to get any better than that, clearcoat is a subtle effect.  We should also check if you have bright HDR reflections in as many engines as possible.

cx20:
@emackey I have confirmed the updated model. I think the difference is easier to understand than when it is flat.
|Babylon.js|Filament|
|:----------:|:----------:|
|[image](https://rawcdn.githack.com/cx20/gltf-test/3547eae63dc3ce5ceb59b346f83bc0a7aa710297/examples/babylonjs/index.html?category=tutorialModels&model=TextureTransformMultiTest&scale=1&type=glTF)|[image](https://rawcdn.githack.com/cx20/gltf-test/3547eae63dc3ce5ceb59b346f83bc0a7aa710297/examples/filament/index.html?category=tutorialModels&model=TextureTransformMultiTest&scale=1&type=glTF)|

PR #268, opened 2020-08-03
Test model for KHR_materials_specular and KHR_materials_ior

proog128:
This is a sample model for KHR_materials_specular (KhronosGroup/glTF#1719) and KHR_materials_ior (KhronosGroup/glTF#1718). It also makes use of KHR_materials_clearcoat and KHR_materials_transmission.


cx20:
I experimentally added this model to [gltf-test](https://cx20.github.io/gltf-test/) and tested it. The display results of some WebGL libraries are as follows.
I don't understand the glTF extension of this model correctly, so I don't know which library shows the correct result.

|Library|Result:|
|:---------------:|:---------------:|
|[three.js](https://cx20.github.io/gltf-test/examples/threejs/index.html?category=tutorialModels&model=ToyCar&scale=10&type=glTF)|image|
|[Babylon.js](https://cx20.github.io/gltf-test/examples/babylonjs/index.html?category=tutorialModels&model=ToyCar&scale=10&type=glTF)|image|
|[Filament](https://cx20.github.io/gltf-test/examples/filament/index.html?category=tutorialModels&model=ToyCar&scale=10&type=glTF)|image||
|[PlayCanvas](https://cx20.github.io/gltf-test/examples/playcanvas/index.html?category=tutorialModels&model=ToyCar&scale=10&type=glTF)|image|

proog128:
Thanks! Babylon.js looks very close to our results:

stellarium_image_43

donmccurdy:
three.js hasn't implemented these extensions yet, so I doubt we're correct. 😇

cx20:
@proog128 Babylon.js is close to a path-traced result, but the windshield is not transparent, is that correct? Is transparent the correct answer?

bghgary:
There is no alphaMode specified in the glTF so opaque should be correct.

proog128:
The transparency on the windshield comes from KHR_materials_transmission (KhronosGroup/glTF#1698). The surface is opaque, but it has a transparent material. See [Section about blend modes](https://github.com/KhronosGroup/glTF/blob/a1af8f003042c3d022c90c885248050c4aaec7b4/extensions/2.0/Khronos/KHR_materials_transmission/README.md#blend-mode) in KHR_materials_transmission. Another test model for transmission is provided in #265.

bghgary:
Yup, totally missed that extension.

sebavan:
But the transmissionFactor is undefined so defaulting to 0 according to the spec. @proog128 shouldn t it kind of disable the extension or "kind of" ?

Also I am seeing 6 in the ior extension which sounds a bit high ?

emadurandal:
Hi, there. I'm called by my friend cx20.

@sebavan Hi, IOR 6  is not so unusual in metals.
Some car paint contains metallic composition.

flashk:
Shouldn't the clearcoatFactor be set to 1? The clearcoatFactor is currently unspecified, which means it will be disabled according to the spec.

proog128:
Oh, you are right, the factors are missing. I'll fix that. Thanks!

The car paint is a special case. Usually, an IOR of 6 is too high for dielectrics, and for metals you should use metalness=1. But in the car paint we have a layered mix of dielectric and metal with different colors for each of them. We use an unusually high IOR to increase the contribution of the metal (top layer) over diffuse (base layer).

cx20:
I updated the ToyCar model to the latest version. 

|Model|Result|
|:------:|:------:|
|[Before](https://rawcdn.githack.com/cx20/gltf-test/2913a7675c3f4a864fe2320c617e26766956ecc4/examples/babylonjs/index.html?category=tutorialModels&model=ToyCar&scale=10&type=glTF)|image|
|[After](https://rawcdn.githack.com/cx20/gltf-test/4bc56d9fa08866a840c836bd4c418cc4e29b7584/examples/babylonjs/index.html?category=tutorialModels&model=ToyCar&scale=10&type=glTF)|image|

The windshield has become transparent, but the background behind can be seen through.

|Expected|Current|
|:------:|:------:|
|image|image|

sebavan:
Perfect this i all expected as to completely work we also need to render to texture the none transparent parts which is what we have as a work in progress here: https://github.com/BabylonJS/Babylon.js/pull/8623

@MiiBond is taking care of enabling it in our loader, so we are at the correct state at this point :-)

sebavan:
The only strange part is the clear coat ior on the path tracer.

@proog128 it looks like on the path traced model the ior of 6 is applied to the top layer ??? not the metallic one, I kind of assume looking at the render.

Should it be ?

proog128:
@sebavan With top layer you mean clear coat? No, the clear coat IOR is 1.5. Semantically, the following happens: 
```
fresnel_blend(
  bottom = fresnel_blend(
    bottom = diffuse,
    top = specular,
    ior = 6)
  top = clearcoat,
  ior = 1.5)
```

(There is no "IOR stacking", the IORs do not influence each other).

A few more images, maybe helpful for debugging:

| IOR | CC off | CC on |
|:-:|:-:|:-:|
| 1.5 | toycar_ior_1 5_nocc | toycar_ior_1 5_cc
| 6 | toycar_ior_6_nocc | toycar_ior_6_cc

sebavan:
@proog128 
> (There is no "IOR stacking", the IORs do not influence each other).

This would explain why in Babylon we see such a difference then :-)

Is it part of the spec that they do not influence each other (I might have missed this part) ? so that I ll modify our material accordingly.

Thanks a lot again for your reactivity.

proog128:
@sebavan Ah ok, that explains why the base layer is so much brighter in Babylon. This is what you mean?

The spec doesn't mention IOR stacking, the example code in the implementation sections doesn't use it. Although I like the idea (after all it's more physically-plausible), it will require the material to be set up with "valid" values for the IOR. And this will break use-cases such as the car paint, where IOR is used in a kind-of spec-gloss workflow. I think it will also reduce compatibility to material models in other renderers, especially if they come from spec-gloss or high IOR for metal etc. For this reason I think we should keep IOR stacking for later, when we have a node-based material system with layering in glTF :-)

proog128:
Non-transmissive parts are now separated and branch is rebased (merge conflicts resolved).

cx20:
I updated the ToyCar model to the latest version. 

|Model|Result|
|:------:|:------:|
|[Before](https://rawcdn.githack.com/cx20/gltf-test/4bc56d9fa08866a840c836bd4c418cc4e29b7584/examples/babylonjs/index.html?category=tutorialModels&model=ToyCar&scale=10&type=glTF)|image|
|[After](https://rawcdn.githack.com/cx20/gltf-test/8f8c7f8c8a0ae366bab3b83a21169d3d3284385b/examples/babylonjs/index.html?category=tutorialModels&model=ToyCar&scale=10&type=glTF)|image|

I think the look of the windshield has been improved.

|Expected|Current|
|:------:|:------:|
|image|image|

cx20:
I noticed that a glTF editor called [Gestaltor](https://gestaltor.io/) supports this extension, so I displayed the model to try it out.

|Library|Result|
|:-----:|:------:|
|[Gestaltor](https://gestaltor.io/)|image|

@UX3D-nopper  The [model](https://github.com/proog128/glTF-Sample-Models/tree/KHR_materials_specular/2.0/ToyCar) is up to date, but the windshield doesn't seem to look as expected. Do you have any advice?

|Expected|Current|
|:-----:|:-----:|
|image|image|

romainguy:
Thanks @proog128! Here is the result in Filament:

Screen Shot 2020-09-21 at 9 05 39 AM

Incidentally my changes to support `KHR_materials_ior` and `KHR_materials_specular` have been merged to clgtf so I should be able to show better results for this model fairly soon.

UX3D-nopper:
> I noticed that a glTF editor called [Gestaltor](https://gestaltor.io/) supports this extension, so I displayed the model to try it out.
> 
> Library	Result
> [Gestaltor](https://gestaltor.io/)	image
> @UX3D-nopper The [model](https://github.com/proog128/glTF-Sample-Models/tree/KHR_materials_specular/2.0/ToyCar) is up to date, but the windshield doesn't seem to look as expected. Do you have any advice?
> 
> Expected	Current
> image	image

My team will have a look at it @UX3D-becher

romainguy:
What is the IBL used in all the reference images btw? I'd like to replicate the same lighting.

bghgary:
I think it's this?
http://www.hdrlabs.com/test2/store/archive/PaperMill_Ruins_E
https://github.com/cx20/gltf-test/blob/master/textures/hdr/papermill.hdr

romainguy:
I realized we were still using our option to change the base layer's color when clear coat is enabled (since enabling clear coat creates a transition from IOR=1.5 instead of IOR=1). With this fixed we're closer to the references (still missing support for the IOR/specular extensions though):

Screen Shot 2020-09-21 at 5 07 44 PM

sebavan:
Actually, we still do have this option turned on as well :-) @popov72 Could you see if it is possible to add an option to disable it for GLTF ?

cx20:
@bghgary @sebavan It seems that Babylon.js `v5.0.0 alpha.18` and `20` changed the display result.
Is this an intended change? I think the previous result was closer to the path-traced result https://github.com/KhronosGroup/glTF-Sample-Models/pull/268#issuecomment-669034471.

|Babylon.js + ToyCar(Specular) model result|
|:-----:|
|image|
|image|

Popov72:
What is the mode file used in the screenshot?

It seems it is not the ToyCar from the glTF sample repo as the latter one has no stripes.

cx20:
@Popov72 

> It seems it is not the ToyCar from the glTF sample repo as the latter one has no stripes.

The `ToyCar` model used for testing is the one included in this PR.
https://github.com/proog128/glTF-Sample-Models/tree/KHR_materials_specular/2.0/ToyCar
-> for  `KHR_materials_specular` and `KHR_materials_ior`  extension

This is a different version from the following model, although the name is the same and therefore confusing.
https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/ToyCar
-> for `KHR_materials_sheen`, `KHR_materials_transmission` and `KHR_materials_clearcoat` extension

Perhaps `ToyCar` in this PR is confusing and needs to be named differently (e.g. `ToyCarSpecular`).

Popov72:
Between alpha 18 and alpha 20 we have implemented the KHR_materials_specular extension as described [here](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_materials_specular).

That means `specularTexture` is now giving only the strength of the specular reflection (in the alpha component) and not the color, which is given by `specularColorTexture`. You will need to modify the KHR_materials_specular entries of the glTF file this way to get the same result than before:

Car:
```gltf
                "KHR_materials_specular" : {
                    "specularTexture" : {
                        "index": 4,
                        "texCoord": 0
                    },
                    "specularColorTexture" : {
                        "index": 4,
                        "texCoord": 0
                    }
                },
```
Fabric:
```gltf
                "KHR_materials_specular" : {
                    "specularColorFactor": [0.729, 0.071, 0.078],
                    "specularTexture" : {
                        "index": 8,
                        "texCoord": 0
                    },
                    "specularColorTexture" : {
                        "index": 8,
                        "texCoord": 0
                    }
                }
```

cx20:
@Popov72 Thank you for confirming this. I modified the model manually as you said and got the same display results as before.

@proog128 What do you think is the right thing to do with this model?

proog128:
@cx20 That's a good question. Although not as polished as the other version with sheen, I think it's still a nice test model to check the specular extension and its interaction with the IOR and clearcoat extensions. In particular the combination of specular and clearcoat is not covered by any other test as far as I know.

cx20:
@proog128  Thanks for fixing the file name.
However, it seems that the `specularTexture` is set in duplicate. Could you please check? Probably one of them should be `specularColorTexture`.

https://github.com/proog128/glTF-Sample-Models/blob/c698c733045c203a3f0a7783677b218bedf56bd9/2.0/ToyCarSpecular/glTF/ToyCarSpecular.gltf#L111-L118

proog128:
@cx20 Oops, you are right, thanks. Fixed it.

emackey:
@proog128 This is an old one, but given that we ended up using a different version of the ToyCar ((here)[https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/ToyCar]), would it be OK to close this version?

Do we still need a test model for metallic IOR colors or anything else from this version that isn't in the other version?

proog128:
The ToyCar model in this PR is a better test-case for the interaction of the specular, ior and clearcoat extensions. I am a bit hesitant to abandon it, because it already found bugs in implementations that are not covered by other test models (as far as I know). Some ideas:

* Merge PR as it is, so that we have two versions of the ToyCar model (`ToyCar` and `ToyCarSpecular`).
* Use `KHR_materials_variants` to add an alternative material to the other ToyCar model.
* Close this PR and create a new test model that checks a couple of corner cases with regard to specular, ior and clearcoat extensions. Maybe with simple geometry like spheres and boxes.

Any opinions?

emackey:
> * Close this PR and create a new test model that checks a couple of corner cases with regard to specular, ior and clearcoat extensions. Maybe with simple geometry like spheres and boxes.

I think this might be the best choice, although I'm open to other opinions (@cx20?).  I think having a few sample shapes alongside some samples of the correct answers (where possible) can allow folks to more easily spot problems without a trained eye.  Also it avoids the "car is different colors in different engines" issue seen in the first few shots on this thread.  It would be  better for the sample model to highlight an implementation getting the wrong answer, rather than showing an assortment of different interpretations.

Can we start by making a detailed list of which features / interactions need testing?

proog128:
> Can we start by making a detailed list of which features / interactions need testing?

* Verify that IORs of clearcoat and base layer are independent: same IOR for clearcoat and base does not make the base reflection invisible. Test with something like `roughness = 0.2`, `ior = 1.5`, `clearcoatIOR = 1.5`. Maybe additionally vary the base IOR, like it is done in the (old) ToyCar. Alternatively, use the (old) ToyCar configuration (which should result in a red appearance instead of green).
* Verify that the `min` in `min(f0_from_ior * specularColor) * specular` is correctly applied (only to `f0_from_ior * specularColor`, not to `specular`). Could be something like `ior = 5`, `specularColor = [2, 2, 2]`, `specular = 1`.
* Verify that renderer correctly applies the `maxValue` function. Test with something like `roughness = 0.4`, `baseColor = [1, 1, 1]`, `specularColor = [0, 1, 0]`. In the bad case, the higher the IOR, the more purple the base layer becomes. In the good case, the base layer stays gray. The same test case could be run with `transmission = 1`.

DrX3D:
@proog128 : This PR uses a branch that has conflicts. We are trying to resolve all PRs in this repo in the next week. If it can be resolved in that time, please do so.

proog128:
Based on the discussion above, I'd suggest to close this pull request. I will come up with a simple test model based on [this list](https://github.com/KhronosGroup/glTF-Sample-Models/pull/268#issuecomment-1087481441).

PR #281, opened 2020-12-04
Update the validator & fix Fox model

lexaknyazev:


emackey:
Does Fox.glb need fixing too?  I guess the CI will tell us...

emackey:
Oh it is fixed, it just wasn't obvious from the way GitHub shows diffs.

emackey:
This all looks good, except Travis is broken/hung/slow throughout most of the Khronos repos, so we don't have a CI result.

emackey:
I merged @donmccurdy's #282 in here, so we could see the new CI run.  It appears to have run twice for the same commit, once for "push" and once for "pull_request".  The latter finished in 54 seconds, but the "push" one is still trying to clone the repo, several minutes later.  Is that a fluke?  Should we only allow one build?

emackey:
Well it passed, after 18 minutes.  We can fine-tune in a follow-up PR.  Merging.

PR #270, opened 2020-08-11
SheenChair updated, renamed, Readme added

echadwick-wayfair:
Sheenchair.gltf updated to current conventions in [KHR_materials_sheen](https://github.com/KhronosGroup/glTF/pull/1688), assets renamed to match parent folder including PascalCase, and readme created with screenshots.

PR #284, opened 2020-12-08
Sheen Chair with Variants, redo

emackey:
This is the Sheen Chair with Material Variants from @echadwick-wayfair, manually migrated from an older branch onto a clean branch.  The filenames were tweaked to match the conventions of this repo.

Supersedes #270.


emackey:
Some testing of this was done by @cx20 in #270.  So with the filenames & branching arrangement fixed, I think this is good to go.

PR #285, opened 2021-01-06
Add KHR_texture_basisu version of the FlightHelmet

bghgary:
UASTC used for non-color textures (normal, ORM). ETC1S used for color textures (baseColor).

TODO:
- [x] Put information on how this asset was generated, what ktx2 files are UASTC vs ETC1S, etc.
- [x] Decide if the settings for UASTC/ETC1S is right for this asset

### DIFF
--- a/2.0/FlightHelmet/README.md
+++ b/2.0/FlightHelmet/README.md
@@ -38,5 +38,5 @@ The model in [glTF-KTX-BasisU](./glTF-KTX-BasisU) has been processed with [glTF-
 
 Donated by Microsoft for glTF testing
 
-[![CC0](http://i.creativecommons.org/p/zero/1.0/88x31)](http://creativecommons.org/publicdomain/zero/1.0/)  
+[![CC0](http://i.creativecommons.org/p/zero/1.0/88x31.png)](http://creativecommons.org/publicdomain/zero/1.0/)  
 To the extent possible under law, Microsoft has waived all copyright and related or neighboring rights to this asset.


emackey:
Since there are no fallback textures, this model must include basisu under `extensionsRequired`.

We should update any automated tools to include this when fallback textures are not provided.

bghgary:
> Put information on how this asset was generated, what ktx2 files are UASTC vs ETC1S, etc.

Options for this:
1. Put this in the README.md for the main model.
2. Put a separate README.md for the BasisU version.

I think 2 makes more sense, but I don't think we've done this before.

bghgary:
> Since there are no fallback textures, this model must include basisu under extensionsRequired.

Good point, though we may want to check with @donmccurdy since this was generated with his tool. I'll update it manually for now.

donmccurdy:
Will be fixed in glTF-Transform in next release, thanks! 👍🏻

cx20:
I experimentally added this model to [gltf-test](https://github.com/cx20/gltf-test#format-tests). I think two libraries currently support [this extension](https://github.com/KhronosGroup/glTF/tree/master/extensions/2.0/Khronos/KHR_texture_basisu).

https://github.com/cx20/gltf-test#format-tests
image

UX3D-haertl:
It seems like the count of the last three accessors (accessorTangents, accessorNormals, accessorUVs) is too big and needs to be reduced by 1. If one calculates the byte offset of the last element of those accessors, it exceeds the length of the buffer. Apparently the last vertex does not have a tangent, normal and UV.

donmccurdy:
I'm not seeing anything in the validator report, but will look into it - thanks!

donmccurdy:
Hm sorry, I'm not able to see what you're finding. These interleaved accessors use the last buffer view, which has 436 vertices and a stride of 48, for a total length of 436 * 48 = 20,928. The last accessor is vec2 floats starting at an offset of 40 bytes, and fits within the last 8 bytes of that stride, so the size seems to work out here.

Perhaps you're trying to allocate 436 * 48 bytes _from the starting byte offset of the accessor_, which would exceed the buffer view's length? I have made that mistake before.

UX3D-haertl:
A yes that was the problem. Thanks for the clarification :)

elalish:
Is this blocked by anything? It'd be great to have a KTX2 sample model in the repo when the PR goes out. Would also be nice to an official test asset for those of us demonstrating support.

bghgary:
There are two TODOs in my original comment. Other than that, I don't think so?

elalish:
Well, I'd rather have something imperfectly optimized than nothing. Shall we just document it and then update it later if we find improvements?

bghgary:
I'm okay with that. Can we get some opinions on the documentation? https://github.com/KhronosGroup/glTF-Sample-Models/pull/285#issuecomment-755433887

emackey:
> I'm okay with that. Can we get some opinions on the documentation? [#285 (comment)](https://github.com/KhronosGroup/glTF-Sample-Models/pull/285#issuecomment-755433887)

I think the model's main README should have a section that explains the BasisU flavor.

donmccurdy:
FYI: https://github.com/donmccurdy/glTF-Transform/pull/222

emackey:
Given the decisions made in #292, the new folder here should be called `glTF-KTX-BasisU`.  I think it's still well worth having a KTX2 version of more than one sample model.

bghgary:
I updated the model using the latest KTX tools and glTF-transform and added a new section to the [README](https://github.com/KhronosGroup/glTF-Sample-Models/blob/b3982198d759385b8787f01efe02ec1f91e5e8fb/2.0/FlightHelmet/README.md#khr_texture_basisu). Should be ready to go unless someone has comments.

PR #286, opened 2021-01-17
Add a Box model with URI-encoded spaces in texture names.

emackey:
This adds a textured box sample model, with spaces in the BIN filename, and URI-encoded spaces in three texture names.  Currently this is how the Blender exporter writes the model, and I believe it's legal.  The validator does not issue any messages for this asset.

The spec was debated in some comments in https://github.com/KhronosGroup/glTF/issues/1449.  Alexey [writes](https://github.com/KhronosGroup/glTF/issues/1449#issuecomment-422041710) that RFC3986 allows for both spaces and URI-encoded spaces when the underlying scheme does not treat a space as a reserved character.  For example, a `file:` scheme would allow both, but an `http:` scheme would allow only the `%20` URI-encoded form, as space is a reserved character in the HTTP protocol.  So, this model tests both ways.

VSCode recently got a bugfix related to this, see https://github.com/AnalyticalGraphicsInc/gltf-vscode/issues/200.

Note that in this sample repo, `model-index.json` is generated by a Python script that apparently does not URI-encode names of glTF files.  Does that need to change?

screenshot_large


donmccurdy:
Does this belong here, do you think, or in https://github.com/KhronosGroup/glTF-Asset-Generator/?

bghgary:
For reference, https://github.com/KhronosGroup/glTF-Asset-Generator/issues/451

emackey:
I'd prefer to merge this to the main sample repo.  I think this really should have been one of our earliest test models all along, as desktop implementations commonly forget or misunderstand URI decoding.  Also it has a nice PBR material on a cube with a glTF logo on it.  Perhaps one of the other cubes is redundant now, if you feel there are too many cubes here. 😄

donmccurdy:
Okay, sounds good — maybe a little heavy on cubes at this point haha, but no need to solve that now. 😬

PR #290, opened 2021-01-22
Add Morph Stress Test model.

emackey:
Eight morph targets join forces to push your vertex attributes past the limit.  [Read more...](https://github.com/KhronosGroup/glTF-Sample-Models/blob/578c64f179d2ea6268a5ffd148bba99e256bf3a0/2.0/MorphStressTest/README.md)

https://user-images.githubusercontent.com/1235080/105434239-250dad80-5c29-11eb-9ee5-e0dc663f8956.mp4


donmccurdy:
Very nice example! Fine with me to merge this any time.

Do you think there's some way this could also indicate whether the "NORMAL" morph attributes are being applied? A client could pass all three tests by dropping those attributes, and I'm not sure whether that's something we'd encourage.

emackey:
> Do you think there's some way this could also indicate whether the "NORMAL" morph attributes are being applied?

I haven't really thought through a test that would indicate that.  The normals on this particular test probably don't change visibly enough to determine that, so a different test model might be required there.

I'll merge this one as-is for now.  If anyone has ideas how better to test normals specifically, I'd be curious to hear.

bghgary:
Cool model! I somehow missed this model completely when it was first posted. I wanted to mention that there is now a technique of using texture arrays to pass in morph targets which effectively allows infinite morph targets without being constrained by vertex attributes. You can find out more about the implementation in Babylon.js in [this video](https://www.youtube.com/watch?v=LBPRmGgU0PE).

PR #291, opened 2021-01-29
Rounded Cube sample for KHR_xmp_json_ld extension.

weegeekps:
This adds the Rounded Cube sample to the samples repository for the upcoming [KHR_xmp_json_ld](https://github.com/KhronosGroup/glTF/pull/1893) extension.


cx20:
To be consistent with other samples, it is better to name the sample model file the same as the folder name is a good idea.

|Folder Name     |File Name|
|:---------------|:--------|

Also, to keep things simple, I think it would be better to put the `.bin` file that this model refers to in the same folder.
This is because some viewers This is because some viewers cannot display correctly when using a relative path.
For example, the [Khronos glTF Sample Viewer](https://gltf.ux3d.io/) fails to display the `.bin` file if it is in a different folder.

PR #294, opened 2021-02-16
GlamVelvetSofa model

echadwick-wayfair:
A model to demonstrate the need for colored specular on non-metal surfaces in a PBR metal-rough material, by using KHR_materials_specular. Model also uses KHR_materials_sheen, KHR_materials_variants, and KHR_texture_transform.


echadwick-wayfair:
In some of my materials I am pushing sheenColorFactor beyond 1, e.g.:
```
                    "sheenColorFactor": [
                        3.25,
                        2.859,
                        2.625
                    ],
```
I believe this is failing the validation though `exit code 1` isn't very clear. Does this mean we can't commit glTF files with these values?

lexaknyazev:
> `exit code 1` isn't very clear

Output for the asset is quite specific:
https://github.com/KhronosGroup/glTF-Sample-Models/pull/294/checks?check_run_id=1914395379#step:4:488

> Does this mean we can't commit glTF files with these values?

These values violate the extension schema, so we should fix them.

echadwick-wayfair:
Thanks. 

Unfortunately I can't reproduce the surface qualities of these velvet fabrics without going beyond the normal range. To get this kind of sheen, I need to make the baseColor really dark, and push the sheenColor really bright. 

Color values >1
image

If I set the sheenColor <1, then I have to increase the baseColor to keep the same overall brightness. But this kills the velvet sheen.
image

The photo reference I need to match:
photoreference_champagne

echadwick-wayfair:
> These values violate the extension schema, so we should fix them.

Would it be possible to amend the schema?

lexaknyazev:
One of color factors' usage (not just sheen) is to replace a single-value texture, so such factors must be within `0.0 ... 1.0` range.

Maybe the sheen material from the photo reference cannot be represented using the glTF sheen extension. Since all three values are above `1.0`, it seems to me that the material would need some kind of "sheen strength" scalar value.

/cc @emackey

echadwick-wayfair:
Ah, that makes sense, thanks. Yes, a sheen strength factor would help here.

cx20:
@echadwick-wayfair 

I think the `generator` property of glTF is a bit long. I think the `copyright` property is more appropriate for copyright notation.

As Is:
```
"generator": "(c) 2021 Wayfair, CC BY 4.0. Generated with babylon.js glTF exporter for 3dsmax 2020 v20210115.1, hand-edited in VSCade with glTF Tools extension."
```

To Be:
```
"copyright": "(c) 2021 Wayfair, CC BY 4.0.",
"generator": "Generated with babylon.js glTF exporter for 3dsmax 2020 v20210115.1, hand-edited in VSCade with glTF Tools extension."
```

See below for how to use the `copyright` property.
https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#asset-1

2.0/GlamVelvetSofa/glTF/GlamVelvetSofa.gltf
```diff
@@ -1,7 +1,8 @@
 {
     "asset": {
         "version": "2.0",
-        "generator": "(c) 2021 Wayfair, CC BY 4.0. Generated with babylon.js glTF exporter for 3dsmax 2020 v20210115.1, hand-edited in VSCade with glTF Tools extension."
+        "generator": "(c) 2021 Wayfair, CC BY 4.0.",
+        "copyright": "Generated with babylon.js glTF exporter for 3dsmax 2020 v20210115.1, hand-edited in VSCade with glTF Tools extension."
```
cx20, line 5: The contents of "copyright" and "generator" are reversed.

echadwick-wayfair:
Is there anything else that needs to be done with this asset to resolve the merge? Thanks.

emackey (approved):

Issue #293, opened 2021-02-15
Dedicated test for correct pixel transfer functions


lexaknyazev:
As of today, only four glTF color textures use sRGB encoding:
* `emissiveTexture`
* `pbrMetallicRoughness.baseColorTexture`
* `KHR_materials_pbrSpecularGlossiness.specularGlossinessTexture`
* `KHR_materials_sheen.sheenColorTexture`

All other textures are sampled linearly. We need dedicated test models that would highlight potential implementation errors in that area.

It's important to note, that working with linear 8-bit values in DCC tools could be quite challenging as some editors may misrepresent such content on screen and output show misleading color-picker values.

donmccurdy:
If we create test assets for these cases, I would prefer to use KHR_materials_specular (https://github.com/KhronosGroup/glTF/pull/1719) over KHR_materials_pbrSpecularGlossiness. Fine to have both, too, but the emphasis will be more on the first moving forward.

emackey:
Is this resolved now?

lexaknyazev:
I'd prefer to keep this open until we have similar tests for:
- `KHR_materials_pbrSpecularGlossiness.diffuseTexture`
- `KHR_materials_pbrSpecularGlossiness.specularGlossinessTexture`
- `KHR_materials_sheen.sheenColorTexture`
- `KHR_materials_specular.specularColorTexture`

Although the SpecGloss is deprecated, Sample-Viewer had a bug with it and the sheen color: https://github.com/KhronosGroup/glTF-Sample-Viewer/issues/295.

DrX3D:
This issue has been opened for almost 3 years. Since the repo has moved and there is a request to keep it open until certain things are done, should this be transferred to Sample-Assets? If there is no response, the issue will be transferred to Sample Assets by 17 Nov.

PR #296, opened 2021-02-17
Add TextureEncodingTest

lexaknyazev:
Addresses #293 (without extensions).

Spheres located on the same horizontal line must look exactly the same. So far Babylon and Filament pass, Three and Cesium partly fail (likely they don't set the relevant `pixelStorei` flag).


emackey (approved):
Fantastic test.  We'll leave this open a day or two for comments, then merge.

emackey:
In VSCode's preview window, it looks like Babylon and Filament pass this test, but ThreeJS and Cesium both fail it.

2.0/TextureEncodingTest/README.md
```diff
@@ -0,0 +1,23 @@
+# Texture Encoding Test
+
+## Screenshot
+
+![screenshot](screenshot/screenshot.png)
```
bghgary, line 5: Can we include a screenshot of what it looks like if the ICC profiles are not ignored?

2.0/TextureEncodingTest/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+This asset tests that textures are sampled with correct transfer functions: base color and emissive must be decoded from sRGB while metallic-roughness values must be passed as-is.
+
+Additionally, this asset tests that gamma values and/or ICC profiles embedded with image files are ignored as the glTF spec requires.
+
+For each of three material slots, four sample models are provided that use different representations of the same effective material properties.
+
+* Models from the first (leftmost) column use only JSON-stored factors to define their materials.
+
+* Models from the second column represent the same materials using 1x1 textures with no color space metadata.
```
bghgary, line 17: Do 1x1 textures work everywhere?

lexaknyazev: Yes. It's a power of two: 2<sup>0</sup>=1.

2.0/TextureEncodingTest/README.md
lexaknyazev: Custom gamma values and ICC profiles are more distinct now. Screenshot added.

lexaknyazev:
This test covers only the base spec. We may want to provide similar cases (factor-texture comparison) for all material extensions.

emackey:
> This test covers only the base spec. We may want to provide similar cases (factor-texture comparison) for all material extensions.

That gets into a bit of a gray area between here and glTF-Asset-Generator.  I think just the core test here is a good start by itself.

2.0/TextureEncodingTest/README.md
bghgary: I have this memory that we had to increase the minimum texture size to avoid some hardware that didn't support smaller than 4x4 images, but maybe that's no longer an issue.

lexaknyazev: I wonder how that hardware would handle mips as they have to eventually reach 1x1.

cx20:
I experimentally added this model to [gltf-test](https://cx20.github.io/gltf-test/). I am going to try this model over the weekend.
Interestingly, in the case of Three.js, it is `NG` in Chrome on Windows, but `OK` on iPad.

BTW, it may be a matter of taste, but I prefer the model to be in the center of the Scene, as in the [TransmissionTest model](https://cx20.github.io/gltf-test/examples/threejs/index.html?category=tutorialModels&model=TransmissionTest&scale=3&type=glTF).

lexaknyazev:
Interestingly, PlayCanvas fails the first column (factors) test only for GLB.

> BTW, it may be a matter of taste, but I prefer the model to be in the center of the Scene

Given that apparently not all engines auto-align the camera, I agree. Should the scene be also scaled to fit `-1; 1` box?

cx20:
> Interestingly, PlayCanvas fails the first column (factors) test only for GLB.

That's certainly interesting. @willeastcott could you please investigate this?

[PlayCanvas + TextureEncodingTest.gltf result:](https://cx20.github.io/gltf-test/examples/playcanvas/index.html?category=tutorialModels&model=TextureEncodingTest&scale=0.2&type=glTF)
image

>  Should the scene be also scaled to fit `-1; 1` box?

Since many of the test models in glTF-Sample-Models are scaled differently, I don't think it is important to fit the scales.
BTW, [gltf-test](https://cx20.github.io/gltf-test/) allows the scale to be specified by the URL argument. For example, `&scale=0.2`.

willeastcott:
Hey, @mvaligursky, can you take a look at this when you have a moment?

lexaknyazev:
@willeastcott @mvaligursky
A bit more context for the observed PlayCanvas issue:
- I see it only with the [GLB version](https://cx20.github.io/gltf-test/examples/playcanvas/index.html?category=tutorialModels&model=TextureEncodingTest&scale=0.2&type=glTF-Binary) of the asset.
- Reloading the webpage sometimes fixes the issue.

willeastcott:
@lexaknyazev Weird, I only see this on my Windows 10 laptop:

image

willeastcott:
Ah, actually, I can repro this now on Win10 in Chrome.

cx20:
The environment I tested is:
MacBookAir + Windows 10 (BootCamp) + Chrome 88 + Intel UHD Graphics 617
Reloading does indeed seem to solve the problem. Is it a timing issue of some kind?

2.0/TextureEncodingTest/README.md
bghgary: Not sure. Anyways, if it works, then I'm fine.

lexaknyazev:
@donmccurdy @emackey 
Now that three.js is fixed, should we wait for Cesium before merging this?

donmccurdy:
Has an issue been filed on Cesium yet? I would suggest that we do that, wait 1-2 weeks, and then merge the PR.

Excellent test, thanks!

cx20:
> Has an issue been filed on Cesium yet?

I have reported this issue as it was not yet in the CesiumJS repository.
https://github.com/CesiumGS/cesium/issues/9390

emackey:
Thanks @cx20 and @lexaknyazev.  I don't think the sample repo needs to wait for every engine to deliver a fix.  We can merge now.

cx20:
@UX3D-nopper I tried to view this model with Gestaltor 2021.2.0. 
However, the label does not seem to be displayed correctly. Can you confirm it?

image

UX3D-nopper:
> @UX3D-nopper I tried to view this model with Gestaltor 2021.2.0.
> However, the label does not seem to be displayed correctly. Can you confirm it?
> 
> image

@UX3D-becher Can pls someone from us can have a look at this?

PR #299, opened 2021-03-01
Change ToyCar normal maps from 16-bit to 8-bit

bghgary:
Per discussion from PBR next sub working group. See KhronosGroup/glTF#1561.


lexaknyazev (approved):
Just for better transparency,

As of today, it's impossible to upload a 16-bit DOM image to the GPU without truncating it to 8 bits (web-browsers do it automatically).

The `EXT_texture_norm16` extension, although available on most desktop and some mobile GPUs, supports texture uploads only from raw array buffers or views.

emackey (approved):
Looks good to me too.

PR #302, opened 2021-03-08
Add Transmission Roughness Test model.

emackey:
This model demonstrates interactions between transmission and the (still draft) IOR and Volume extensions.

In this case, Volume is used solely to declare a thick-walled surface, and IOR produces most of the effects shown here.

See [the README](https://github.com/KhronosGroup/glTF-Sample-Models/blob/89e232867742d5c895179eb38e2503eb2d94c467/2.0/TransmissionRoughnessTest/README.md) for more details.


proog128:
Great explanation and great test model, thank you!

PR #305, opened 2021-03-29
Add test object with unicode characters in file, material and mesh names

kovacsv:
I've created a new test object to test if importers can handle unicode characters.


2.0/model-index.json
```diff
@@ -567,6 +568,14 @@
       "glTF": "TwoSidedPlane.gltf"
     }
   },
+  {
+    "name": "UnicodeTest",
+    "screenshot": "screenshot/screenshot.png",
+    "variants": {
+      "glTF": "model-\u2764\u267b-model.gltf",
+      "glTF-Binary": "model-\u2764\u267b-model.glb"
```
emackey, line 576: Did this thing choke our Python script here?  Or was this run before a rename?

scurest: No, Python's `json.dump` writes pure ASCII output by default; any non-ASCII characters are written escaped. If you want to change it, pass `ensure_ascii=False` [here](https://github.com/KhronosGroup/glTF-Sample-Models/blob/0bc2ad7535aba4e16fb1afba1f85fd827bbbea6d/generate-index.py#L37).

kovacsv: What format would you prefer? Escaped unicode characters like now, or the original characters?

kovacsv: Btw ensure_ascii is not enough, the final code should look something like this:
```python
with codecs.open("model-index.json", "w", encoding="utf-8") as f:
    json.dump(index, f, indent=2, sort_keys=True, ensure_ascii=False)
    f.write("\n")
```

emackey: Apologies for the noise here.  Escaped is totally fine.  Somehow this looked broken to me last night, but it's correct.

emackey:
This is a great test model.  My only remaining concern here is the actual texture image being used.  It's funny and tongue-in-cheek, but, when I first saw it my instinct was that Unicode decoding had failed.  It took me a moment to realize the failure was intentionally encoded as a raster image.

I'm hesitant to wreck a good joke, but I'd like to see the model clearly show success where it's due.  Can we do that?

Thanks!

kovacsv:
You are right, I've had the same concerns about this joke, because it can cause painful times for programmers looking for the issue :) I've replaced the texture with a new one that shows the used characters as an image.

lexaknyazev:
This asset uncovers an issue with our JSON-Schemas. Specifically, `"format": "uriref"` that is used there wasn't defined until draft-5, so it has no effect right now. Enforcing `uriref` (already renamed to `uri-reference` in draft-6) would make this asset invalid.

The solution seems to be using `iri-reference` that allows non-ASCII symbols.

cx20:
I agree that this model is a good test model. 
A little worrisome is that the file name rules are different from other models.
For example, in the case of Duck, the folder name and model name match as shown below.

|Folder Name     |File Name|
|:---------------|:--------|

If you follow this method, I think it will be as follows.

|Folder Name     |File Name|
|:---------------|:--------|
|`Unicode❤♻Test`/glTF       |`Unicode❤♻Test`.gltf|
|`Unicode❤♻Test`/glTF-Binary|`Unicode❤♻Test`.glb |

I don't think it's a rule that you have to do that, but if it's not, it's just a bit of a hassle for my [gltf-test](https://cx20.github.io/gltf-test/) because I have to write individual if statements.

kovacsv:
Good point @cx20, I've renamed the folder and now everything follows the proposed convention.

cx20:
@bghgary This model looks fine in Babylon.js, but doesn't seem to look right in the Windows `Paint 3D` or `Explorer previews`. Can you confirm it?

Paint 3D result: The texture is not displayed.
image

Explorer preview: Not previewed
image

kovacsv:
@cx20 Also, PowerPoint doesn't open it while there are unicode characters in the file name. After renaming it to `model.glb` PowerPoint works fine, and there is a preview in explorer (but without textures).

emackey:
> The solution seems to be using `iri-reference` that allows non-ASCII symbols.

@lexaknyazev Is this a breaking change to glTF 2.0?  Or is it something we could patch?

scurest:
Would also be good to test percent encoding (from [Relationship between IRIs and URIs](https://tools.ietf.org/html/rfc3987#section-3.1) from RFC3987), eg.

```
"uri": "Unicode%E2%9D%A4%E2%99%BBBinary.bin"
```

This is a valid RFC3986 URI so it should be valid according to current spec.

cx20:
> Also, PowerPoint doesn't open it while there are unicode characters in the file name.

@kovacsv PowerPoint on my PC seems to be able to insert this model without any problems. Excel was fine too.

Windows 10 Version 20H2 (OS Build 19042.867) + Office Pro Plus 2019 Version 2102 (Build 13801.20360)

image

The preview in explore was similar to yours. It was displayed when renamed to `model.glb` (but the texture is not displayed)

image

bghgary:
@cx20 I see the same as what you see, for Windows explorer preview pane (after rename), Paint 3D, PowerPoint, etc.

cx20:
@bghgary Thank you for your confirmation. I am looking forward to it being fixed someday.

bghgary:
Don't hold your breath on this one. It might be a while.

emackey:
@lexaknyazev @vpenades Given where your discussion in https://github.com/KhronosGroup/glTF/issues/1978 ended up, is this asset valid and good to merge as a sample model?

emackey:
Maybe we could take this one as-is for Unicode, and have a separate PR for a model that tests escaped URIs?

vpenades:
@kovacsv  you can have a single file with multiple URIs, each one testing a different aspect of character encoding.

@emackey you can have some tests for each specific case, but I would also have a model testing all in a single string. Consider a decoder that is able to handle only one use case at a time, but not both at the same time.

emackey:
I agree it would be good to test both together.  I guess I was thinking of these more along the lines of unit tests, which these models are often used for.  In those cases, each model gets a simple pass/fail indicator, and I was imagining one pass/fail for Unicode, and another pass/fail for escaped strings (particularly escaped spaces in the URI needs testing).

But you're right, fewer models is better, and a combo test is also good.  Perhaps the ideal solution might be a single model with a row of test tiles:  One tile tests Unicode by itself, another tests escaped spaces, another for escaped strings with Unicode mixed in, etc.

emackey (approved):

emackey:
Thanks @kovacsv, apologies for the long delay here.

PR #307, opened 2021-04-12
Fix link for license

echadwick-wayfair:
The original link (http://research.scea.com/scea_shared_source_license.html) points to an inactive website. The last time it was active according to the Wayback Machine was March 2016 (https://web.archive.org/web/20160320123355/http://research.scea.com/scea_shared_source_license.html).

### DIFF
--- a/2.0/Duck/README.md
+++ b/2.0/Duck/README.md
@@ -9,6 +9,6 @@ Copyright 2006 Sony Computer Entertainment Inc.
 
 Licensed under the SCEA Shared Source License, Version 1.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at:
 
-http://research.scea.com/scea_shared_source_license.html
+https://web.archive.org/web/20160320123355/http://research.scea.com/scea_shared_source_license.html
 
 Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.


javagl:
Thanks for this fix! (We might also consider just adding the contents of the license as a separate license file in the respective subdirectory).

Before merging: Could you also fix the link in https://github.com/KhronosGroup/glTF-Sample-Models/tree/108cde796df9d52858e2884e26b90f4b96c0f0c9/1.0/Duck ? (If not, I can do it later, after merging this PR)

echadwick-wayfair:
OK thanks, I've edited the link in that branch and merged it. Closing this redundant branch.

echadwick-wayfair:
Oops, that link was for 1.0, this pull is for 2.0. I will merge this PR now to update 2.0

echadwick-wayfair:
Ah, authors can't merge their own PRs, make sense. I'm new here, haha. Thanks for the review.

javagl:
Sorry, I'm a bit confused: This PR still only fixes the link in the `2.0` directory - did you open another one for the `1.0` directory? 
(If not, I'd merge *this* PR, and then open+merge another one for the `1.0` directory)

echadwick-wayfair:
[You posted](https://github.com/KhronosGroup/glTF-Sample-Models/pull/307#issuecomment-817895110) a link to the 1.0 pull request. However my PR is for the Duck in the 2.0 section. It seems odd to me we have two copies of essentially the same model (albeit with different subfolders).

javagl:
The link that I posted was not to a PR, but just to the 1.0 version of the duck. Maybe to clarify: The "Duck" model and the README with the wrong link appears two times, namely in the directories

    glTF-Sample-Models/1.0/Duck/README.md
    glTF-Sample-Models/2.0/Duck/README.md
    
The second one, with `2.0` in the path, is the one that you fixed with this PR. 
The first one, with `1.0` in the path, still contains the old link (and I'll fix that now).

The model appears two times, because the first one is a glTF with version 1.0, and the second one is a glTF with version 2.0. (Version 1.0 is basically no longer used today, but it's still part of history, and I'll also fix the link there).

javagl:
Sorry, I've just been confused a little. The fix for the `1.0` path apparently went directly into `master`? However, now both links seem to be fixed.

> It seems odd to me we have two copies of essentially the same model (albeit with different subfolders).

One is a glTF with version 1.0, and the other one is a glTF with version 2.0. But maybe that wasn't what you meant...

PR #310, opened 2021-05-04
Specular Test

emackey:
See the [README](https://github.com/KhronosGroup/glTF-Sample-Models/blob/specular-test/2.0/SpecularTest/README.md).


abwood (approved):
Excellent writeup! Thanks @emackey

PR #331, opened 2021-10-05
Add sample glTF for KHR_materials_iridescence extension

PascalSchoen:


cx20:
Sorry for the minor point.
The current `screenshot.png` image seems to be a bit large in size.
I would recommend renaming the current `screenshot.png` to `screenshot_large.png` and preparing a new, smaller screenshot.

For example, the following may be helpful.
https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/MosquitoInAmber/screenshot

The reason for this is that the screenshot may also be used to display thumbnails, so if it is too large, it may take a long time to load the entire page.

2.0/IridescenceSuzanne/glTF/IridescenceSuzanne.gltf
```diff
@@ -0,0 +1,315 @@
...
+        "generator" : "Khronos glTF Blender I/O v1.1.45",
+        "version" : "2.0"
+    },
+    "extensionsRequired": [
+        "KHR_lights_punctual",
+        "KHR_materials_ior",
+        "KHR_materials_transmission",
+        "KHR_materials_volume",
+        "KHR_materials_iridescence"
+    ],
```
emackey, line 12: None of these extensions ever belong in `extensionsRequired`, because none of them are needed to understand the basic geometry & file format of what's here.  Viewers can skip newer features and still load valid data from the core glTF format without these extensions.

"Required" extensions are things like Draco (when no fallback is present) and KTX2 (no fallback).

emackey (commented):
Thanks for this sample @PascalSchoen.  This repo has a number of abstract copies of Suzanne and cubes and such, but I'm wondering if for the PBR extensions we could lean on more real-world kinds of objects?  An insect with iridescent wings, or a bubble-maker machine with animated soap bubbles coming out of it, or something.  It would give the viewer something more recognizable to see when they view the effect in a capable viewer.  Alternately, create an abstract test model that puts the different fields to the test, showing different IOR swatches, different film thicknesses, etc.  That can be useful for conformance testing.

2.0/IridescenceSuzanne/glTF/IridescenceSuzanne.gltf
PascalSchoen: @emackey I added two more examples to this PR which show iridescence effects on metallic/dielectic materials with varying parameters for (complex) material IOR, thin-film IOR and thin-film thickness.

echadwick-wayfair: @PascalSchoen these models are great to have, I especially appreciate the 3-parameter grids of spheres. 

Could you explain the complex IOR parameter? How are those values being applied in the glTF materials? I can't seem to find them in IridescenceMetallicSpheres.gltf. 

Also it seems that KHR_materials_ior is not present in any of the materials in IridescenceMetallicSpheres.gltf, although it is defined in extensionsUsed.

PascalSchoen: @echadwick-wayfair The complex IOR is calculated from the base color of the material using the artistic IOR function from this paper: http://jcgt.org/published/0003/04/03/paper.pdf

I added this to the implementation in the glTF Sample Viewer: https://github.com/PascalSchoen/glTF-Sample-Viewer/blob/feature/KHR_materials_iridescence/source/Renderer/shaders/functions.glsl#L87

It is completely debatable if this is the way to go. I just tried to be as close as possible to other implementations like in RenderMan without introducing another parameter to the KHR_materials_iridescence extension. Usually you would require a reflectivity (base color/F0) and edge tint parameter. I approximated edge tint with the base color for metals. By using this algorithm we can have three different complex IOR values - one for R, G and B.

emackey:
This model needs a README, similar to other models here.  What license goes on this?  Can it be CC0, or CC-BY 4.0?

PascalSchoen:
@emackey I added a README to all the models with CC0 license.

cx20:
There seems to be an artifact in the head of the rightmost Suzanne  (transparent Suzanne) in the IridescenceSuzanne model. Is this a problem with the model?

https://cx20.github.io/gltf-test/examples/babylonjs/index.html?category=tutorialModels&model=IridescenceSuzanne&scale=1.0&type=glTF

image

PascalSchoen:
@cx20 yes, the UVs are not setup properly here. The Suzanne was exported from Blender and if you were to create one there and use it's initial UV coordinates, you get the same effect.

PR #335, opened 2021-11-15
Update SimpleSkin to place the parent bone at the origin.

emackey:
@javagl This is my attempt at fixing https://github.com/KhronosGroup/glTF-Tutorials/issues/44.

Currently this model's two bones share the same location.  This is visible by importing into Blender, although the orientation of both nodes is shown incorrectly.  Blender can't correctly guess the bones' primary axis, due to them being colocated.

https://user-images.githubusercontent.com/1235080/141862374-63e09c95-c54c-4460-91d6-b4f3cde1d0bd.mp4

This PR moves the `translation` and its associated `inverseBindMatrix` such that one of the two nodes is positioned at the origin.  Blender's import gets this one correct, because it presumes the child has moved to the end of the parent bone.  But more importantly, the parent bone really is placed at the origin now.

https://user-images.githubusercontent.com/1235080/141862574-ea4cf0e3-1f71-4c85-8c29-7fd824dc63c6.mp4


emackey:
Here's a visual diff of the `.bin` file before and after.  One `-1.0` has been changed to `0.0`.

bin diff screenshot

emackey:
Of course, I started from the sample-model default branch here, not @javagl's latest.  So, this probably can't be merged as-is, and would need to be rebased on @javagl's latest work.

javagl:
(Sorry, I had written my latest comment in the linked PR before I saw this PR here). I still wonder how it can be justified that the root/parent node does not have a translation. I.e. if the parent bone had a length of 2.0, would a translation of (0,2,0) be stored in the *child* node? That seems strange...

javagl:
Based on the dicussion in the linked issue, I *think* that this could be merged. (Or I could create a similar PR, and update the data layout images at https://github.com/KhronosGroup/glTF-Sample-Models/tree/a01b42951ed199597936330ada30bfa4a41c2d05/2.0/SimpleSkin#data-layout accordingly...)

emackey:
@javagl Yes, it looks like https://github.com/KhronosGroup/glTF-Tutorials/pull/64 and this PR are in agreement currently.

Do you want to merge this one?  Or should we hold for an updated diagram?

javagl:
I'll to the final wording tweaks in the other PR and post a notification when ready (not later than during the weekend). Creating the updated diagrams is simple (as in `writeDiagrams(createSimpleSkinData())`), but I'll probably check and compare the diagrams and data several times... 
Depending on what you prefer, I could just add them to this PR branch, or branch off from this one and create a PR with the diagrams added. (Merging this and then creating a PR for the diagrams only would be possible, but not as clean).

emackey:
If you can just add them to this branch, that sounds easiest.  Thanks!

emackey:
Thanks @javagl!

Issue #262, opened 2020-05-24
Sponza model normal maps have incorrect Y direction


aejsmith:
glTF specifies that normal maps have Y as up, however the normal maps for this model have Y as down.

My engine currently does not use precomputed tangents from the model, instead it always calculates the tangent basis in the pixel shader. When doing this, I get upside down normals for the Sponza model, so I have to manually invert Y to fix it.

Example textures where it's particularly obvious:
https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/2.0/Sponza/glTF/759203620573749278.jpg
https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/2.0/Sponza/glTF/13982482287905699490.jpg

emackey:
Generally I agree.  The [spec for `normalTexture`](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#materialnormaltexture) is fairly clear on this.

That said, I'm not sure it's legit for readers to ignore the embedded tangents and compute their own.  But that doesn't make it OK for models to flip the Y axis in the normal map.

emackey:
Fixed in #336.

PR #336, opened 2021-12-20
Fix Sponza normal maps (Y+ up)

IcanDivideBy0:
As described in #262, normal maps for the sponza model use the wrong y direction: [the doc](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#_material_normaltexture) makes it pretty clear that normal maps should be Y+ up.
I wish I found the issue before spending hours debugging my shaders :(

Then it was just a matter of running `convert $f -channel G -negate $f` on every normal maps.


donmccurdy:
Thanks @IcanDivideBy0! Do your shaders use the included vertex tangents in this model, or are they doing something else? I think it's the case that the vertex tangents were wrong for the inverted normal maps, and will be correct with your fix, but just want to double-check that I don't have that backward... I can help with flipping the tangents if needed. /cc @emackey

IcanDivideBy0:
@donmccurdy yes, i'm using the provided tangents in my shaders, but, I've imported the gltf in blender and re-exported as glb... so i don't use directly the gltf file of this repository, although I dont think blender recomputed them

donmccurdy:
Blender will indeed recompute the tangents – it can't store imported tangents within a Blender workspace as far as I know. The correct way of computing tangents does not depend on the normal map itself though, so this change should be OK assuming the originals were not flipped to begin with.

emackey:
Hi, thanks @IcanDivideBy0.  I believe this PR does require a change to the exported tangents, not just the images.  Blender is doing the flip because it doesn't store glTF's tangents, but the raw model here would need the tangents fixed if we took the flipped images.

Perhaps the sample model here could be replaced by one that's been round-tripped through Blender.  Would anyone object to that?  Overall I think it would be great to have this model follow the proper glTF normal map convention.

emackey (changes requested):
This can't be merged as-is, because the tangents included with the model are aligned to the original -Y normal maps, not these new +Y normal maps.

donmccurdy:
If it's just a matter of inverting `tangent.w`, we can do that in https://gltf.report/ by loading the model and running the script below. Running it through Blender could have more potential side-effects, like welding/unwelding vertices.

```ts
import { Primitive } from '@gltf-transform/core';

let primCount = 0;

function invertPrimTangents(prim: Primitive) {
	const tangent = prim.getAttribute('TANGENT');
	if (!tangent) return;

	const el = [];
	for (let i = 0, il = tangent.getCount(); i < il; i++) {
		tangent.getElement(i, el);
		el[3] *= -1;
		tangent.setElement(i, el);
	}

	primCount++;
}

for (const mesh of document.getRoot().listMeshes()) {
	for (const prim of mesh.listPrimitives()) {
		invertPrimTangents(prim);
	}
}

console.debug(`Inverted tangents of ${primCount} primitives.`);

```

IcanDivideBy0:
@donmccurdy I've run your script in gltf report, now a straight import in blender seems to produce correct normal mapping. Didn't know about gltf.report, looks like a very handy tool, thanks!

emackey (approved):
Yes, awesome.  Thanks so much @IcanDivideBy0 and @donmccurdy, this model has relied on inverted tangents for a long time.

PR #339, opened 2022-01-21
Remove "Skins" from Virtual City's description

StoneWolf:
Removed "Skins." from Virtual City's description, as it does not have skins.


javagl:
Right after hitting "Merge", I noticed that the README for the 1.0 models at https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/1.0 also mentions "Skins" there. (I'll not that as an internal TODO).

PR #340, opened 2022-03-11
Emissive Strength test model

emackey:
screenshot_large_bloom

/cc https://github.com/KhronosGroup/glTF/pull/1994


cx20:
I have experimentally tried to display this model in [gltf-test](https://cx20.github.io/gltf-test/).
The following is the result of the display in Babylon.js.

https://cx20.github.io/gltf-test/examples/babylonjs/index.html?category=tutorialModels&model=EmissiveStrengthTest&scale=0.3&type=glTF
image

emackey:
Looks great! Note that any implementation showing a white box gets a pass for this test.  This extension does not mandate a bloom effect around the cubes.  Thanks!

cx20:
@emackey 

> Note that any implementation showing a white box gets a pass for this test. 

The following is a display example in Filament. There is no bloom effect, but it is displayed as a white box, so is it safe to assume that the test has passed?

https://cx20.github.io/gltf-test/examples/filament/index.html?category=tutorialModels&model=EmissiveStrengthTest&scale=0.2&type=glTF
image

emackey:
Hmm, yes, the extension is certainly supported. Odd that the middle box turns white, the linear rgb there should be `vec3(0.1, 0.5, 0.9) * 4.0` meaning the red channel should be 0.4 prior to conversion to sRGB.  So I don't understand how the middle cube came to be full white, we'll have to ask the Filament team.

cx20:
@abwood When I view a model with `KHR_materials_emissive_strength` extension in Filament, the color seems to be different than expected. Any advice?

abwood:
I suspect it might have to do with differences in exposure and/or tone mapping settings between these viewers. I'll continue to investigate as well.

abwood:
Could tone mapping in the filament viewer be set to "linear"? Using the gltf_viewer sample in Filament, I get the following result:
image

Looking at a captured frame in RenderDoc, I can confirm the incoming emissiveFactor for that middle box is what we would expect (from @emackey, red channel should be 0.4):

image

Given the other carefully selected parameters (basecolor, metallic, and roughness), I believe the emissive is all that should be coming through on this model.

Continuing to walk through RenderDoc, indeed these colors are rendered to a half float texture:
image

Finally, it looks like the colorGrading render step shifts the colors towards the final output:
image

Summary:
I'm a bit more confident that this is a color grading, tone mapping discrepancy. Also, Renderdoc is an amazing tool.

emackey:
Great investigation @abwood.  I agree Filament has done the correct thing with this extension, and just has their own tone mapping / exposure calculations going on downstream.

@cx20 This can be marked as "pass" for Filament.  Thanks!

abwood (approved):

PR #347, opened 2022-04-14
Punctual lights example model

mlimper:
This PR adds a "nicer" example model for the punctual lights extension:
https://github.com/DGG3D/glTF-Sample-Models/tree/master/2.0/LightsPunctualLamp

@teresavieggas


emackey:
Can you delete the un-optimized versions here, and just make the "optimized" ones official?  We're trying to keep the size down as much as possible.  Thanks!

lexaknyazev:
> We're trying to keep the size down as much as possible.

Just a drive-by comment: to actually reduce the size, the PR's history would need to be rewritten.

lexaknyazev:
Besides, it looks like "optimized" and "unoptimized" models are different in terms of materials and used textures.

@mlimper please confirm what's correct here.

mlimper:
Hey, thanks to both of you - very valuable feedback! :-)


I believe the original model was only included here to illustrate a bit better how the modeling process went (original -> optimized model), but it's definitely not a must to include that one... especially having learned about the issue of a "bloated" sample repository. Also, as you noticed, both versions, unoptimized and optimized, are using differently arranged materials (non-atlased vs. texture atlas), but what's most important here is the use of the punctual lights extension, and that should be similar for both... so, the optimized one will do the job as well.


So we'll just keep the _optimized_ one and refer to it as "glTF" / "glTF-Binary" (instead of "glTF-Optimized-..."). While doing so, we'll also re-do the history accordingly (squashing some commits on the way) ... thanks again, we'll then ping you here as soon as the PR is ready for another review!

PR #353, opened 2022-05-19
Test models for KHR_animation_pointer

hybridherbst:
**Animation pointer sample models.**
[Link to the KHR_animation_pointer spec proposal](https://github.com/ux3d/glTF/tree/extensions/KHR_animation_pointer/extensions/2.0/Khronos/KHR_animation_pointer)

- Animate All The Things (wide coverage of KHR_animation_pointer examples)
- Emissive Fireflies (simple emissive intensity animation sample)
- Animated Waterfall (rough time-of-day cycle, UV animation)
- Dragon Material Animation (showcases advanced glTF material features and animates them)

**Supported Viewers**

Can be viewed in Babylon Sandbox, Gestaltor, Needle Engine or with [our three.js fork](https://three.needle.tools/examples/?q=loader_mu#webgl_loader_multiple). 

**Videos**

https://user-images.githubusercontent.com/2693840/200132202-487d7524-6067-48df-a102-1fcf7c870c78.mp4

https://user-images.githubusercontent.com/2693840/169412149-a1eafb07-1a71-489b-a409-18929b8b786b.mp4

https://user-images.githubusercontent.com/2693840/169412154-394c48e1-4dca-461f-8d62-843ae951a535.mp4

https://user-images.githubusercontent.com/2693840/171230924-875cb876-a151-4c03-bc0e-a79c7856a37d.mp4


emackey:
@lexaknyazev Are the validation errors here expected?

hybridherbst:
I believe the main error here is `Animation channel has the same target`, which I think happens because some samplers still (incorrectly) have a Node target set. I've fixed that in the exporter but need to re-export the models.

lexaknyazev:
@hybridherbst you can ignore `ANIMATION_DUPLICATE_TARGETS` errors for now, since this validation rule does not check for extensions.

hybridherbst:
Yep, I figured - I rather meant "this probably needs to be adapted in the Validator once the extension has been ratified" :)

UX3D-nopper:
@hybridherbst Can you please update your examples?

The property inside `KHR_animation_pointer` is `pointer` and not `path`

https://github.com/KhronosGroup/glTF/blob/9ec5cfec24be5e6e6c24619769216953223ac24b/extensions/2.0/Khronos/KHR_animation_pointer/README.md#examples

hybridherbst:
Thanks @UX3D-nopper for updating the sample in the Readme :)
I updated the models now, deprecated two old ones and added glTF for the DragonAttenuation and AnimateAllTheThings samples.

UX3D-kanzler:
There is an issue with the animation pointer in the Fireflies asset:
Material pointer is `/materials/0/baseColorFactor` but it should be `/materials/0/pbrMetallicRoughness/baseColorFactor` instead (same for material 1 and 2).

In the Waterfall asset, the following pointer is used:
`/materials/2/baseColorTexture/extensions/KHR_texture_transform/scale`
Here, too, the pbrMetallicRoughness path prefix is missing.
-> `/materials/2/pbrMetallicRoughness/baseColorTexture/extensions/KHR_texture_transform/scale`

The DragonAttenuation-MaterialAnimation has no animation at all.

UX3D-nopper:
Updated the specification, when the pointer is valid/invalid.

hybridherbst:
@UX3D-nopper the models you mentioned should be fixed, thanks again.

UX3D-kanzler:
I've found another issue: `/materials/{}/emissiveFactor` must be VEC3 but in the assets an accessor of type VEC4 is referenced.

Furthermore, there is an animation pointer with value`/materials/0/extensions/KHR_materials_emissive_strength/emissiveStrength` in Fireflies.gltf but the materials do not contain the `emissiveStrength` property.

pandaGaume:
Hello,
Thank's for the models.
I've found that camera angles animation are set as degree into the AnimateAllTheThings sample, where it should be radian.
Agree with @UX3D-kanzler that emissiveFactor is set to VEC4 and that some extensions missing. This last point may open to discussion on what should be the behavior of the reader when the extension is not present (undetermined yet, nicely failed or adding the extension ??).
One last, but i may be wrong, it's appear that bare value into the UV animation seems Y inverted. 
Again thank's for your work.

hybridherbst:
Thanks @UX3D-kanzler and @pandaGaume for the comments –

I updated the models with
- VEC3 instead of VEC4 for `emissiveFactor`
- all materials now properly specify extensions if any values in those are animated
- all files now properly specify animated values in `extensionsUsed`
- camera `fov` in radians instead of in degrees, good catch @pandaGaume

UX3D-kanzler:
Thank you for the update, I really like these models. Two more comments:
- The color accessor for punctual lights should be VEC3 instead of VEC4 as well.
- In your short clip, the grass in the waterfall model looks animated but I no longer see this animation in the three.js viewer.

hybridherbst:
@UX3D-kanzler both issues should now be fixed. Punctual lights now are VEC3, and the missing keyframes have been added again - it was an issue with the file (too few keyframes exported) not with the viewer. 

It exposed a fun bug in the exporter too - keyframe reduction was only ever looking at previous and next keyframes and not the last exported, so could miss entire animations when the values changed below the detection threshold. Thanks!

pandaGaume:
Hey guys, 
just coming back with the still pending question. Everything is working fine, but the UV offset where i have to use 
`(U,1-V)`
 to get a correct result from the waterfall sample. 
I just re-use the same code we using for the `KHR_texture_transform` which is working fine with the [official sample](https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/2.0/TextureTransformTest/glTF/TextureTransformTest.gltf#L76).
Any thoughts ???

hybridherbst:
An update from my end:
- marked the PR as ready for review
- added explanation labels to the Animate All The Things sample
- added Readmes
- moved the files to the toplevel structure.

https://user-images.githubusercontent.com/2693840/200132182-70e17630-d510-4c24-835a-7f34a8bdd52b.mp4

@emackey maybe a new section in the main Readme makes sense for KHR_animation_pointer models since that is quite a big extension? Also, I wasn't sure if I'm supposed to already add the models to the model list json in the PR or if that's happening later.

Hackn0214:
I made testing data for KHR_animation_pointer.
Upload it if useful.

https://github.com/KhronosGroup/glTF-Sample-Models/assets/60141040/3ce2ca5e-ea6d-4fa3-8b1a-6f16f2c0fba9

DrX3D:
@emackey , @hybridherbst @dwrodger : Has this model been checked for accuracy and correctness against the latest AnimationPointer extension? Please respond by 29 Nov so this PR can be correctly handled.

dwrodger:
> @emackey , @hybridherbst @dwrodger : Has this model been checked for accuracy and correctness against the latest AnimationPointer extension? Please respond by 29 Nov so this PR can be correctly handled.

I don't believe anything has changed in the latest extension that would alter whether content is correct or not, it has only changed where the list of readable and writeable properties are listed and where the concepts are defined.
@lexaknyazev may have more details, but I believe that if this content was good before our changes, then it is still good after our changes.

DrX3D:
Closed issue so repo can be archived. An [issue])https://github.com/KhronosGroup/glTF-Sample-Assets/issues/82) was created in Sample Assets to maintain connection.

PR #354, opened 2022-05-24
Squash-commit of optimized LightsPunctualLamp model from DGG.

emackey:
This is a continuation of #347.

@mlimper @teresavieggas Are you OK with this?


mlimper:
Oh thanks Ed, this is very kind - yes, fine with this!

cx20:
Thank you for providing a great model.
BTW, It may be a matter of taste, but I thought it would be easier to compare and understand if the lamp could be turned on and off like the [StainedGlassLamp](https://cx20.github.io/gltf-test/examples/babylonjs/index.html?category=tutorialModels&model=StainedGlassLamp&scale=3&type=glTF).

emackey:
@cx20 Yes, but that doesn't quite work here with the use of KHR_lights_punctual, which cannot be affected by material variants.  There's an upcoming "animation pointer" extension that might be able to animate the brightness of a punctual light (to turn it on or off), but that's still in the draft phase for now.

PR #360, opened 2022-10-05
Add AudioSimple asset

robertlong:
This PR contains a minimal sample asset for testing the draft [KHR_audio](https://github.com/KhronosGroup/glTF/pull/2137) extension.

It includes the source model and a glTF + glb version of the asset. The model was generated in Blender and the audio extension was added by hand. The glb was compiled using Third Room's implementation of glTF-transform.


2.0/AudioSimple/README.md
```diff
@@ -0,0 +1,21 @@
+# Audio Simple
+
+Simple asset to test the KHR_audio extension. This model contains one global audio emitter and one positional audio emitter. The positional audio emitter is attached to the node with the blue sphere mesh. This sphere is animated around the origin of the scene.
+
+In a correct implementation of KHR_audio, you should hear the talking audio coming from the blue sphere circling around the origin. You should also hear some ambient audio including tropical birds chirping at a constant 50% gain no matter where your audio listener is in the scene.
+
+## Screenshot
+
+![screenshot](screenshot/screenshot.jpg)
```
cx20, line 9: Since the file extensions do not match, you probably need to change the extension of the referenced file to '.png'.

cx20: Also, I think the size of `screenshot.png` is bigger than the others. can you change `1200x1200` to about `400x400`? 
This is because `screenshot.png` is also used for list display.

javagl: @cx20 On the one hand, I agree. On the other hand, it's a bit subtle:

- One possible reason to use smaller screenshots is to not bloat the size of the repository. This _particularly_ refers to samples where the _visual presentation_ is not important (like this one): There is no _visual detail_ that is relevant in the screenshot. One could argue about that for other models, where the screenshot shows "a nice rendering result"
- The size of the screenshot affects the column in the table in a non-obvious way. Markdown does not _explicitly_ allow setting the column width for a table. And the GitHub markdown preview apparently does some sort of "weighting": The screenshot here is 1200 pixels wide, but the table column is not - it is only "relatively wide" compared to the other columns...

Subjectively, I like such screenshots and their width to be a bit more "regular" and "predictable". For example:

- This is the current "Showcase" table: https://github.com/javagl/glTF-Sample-Models/tree/master/2.0#showcase-1 
- This is the "Showcase" table with fixed image widths: https://github.com/javagl/glTF-Sample-Models/tree/table-column-width/2.0#showcase-1

The latter can (unfortunately) only be achieved with some (non-markdown) `` tags. But I think that it looks nicer this way...

javagl: (It's a detail, but I just brought this up in https://github.com/KhronosGroup/glTF-Sample-Models/issues/362 nevertheless...)

robertlong: What's the final decision on this? I fixed the file extension

2.0/AudioSimple/glTF/AudioSimple.gltf
```diff
@@ -0,0 +1,381 @@
...
+      "scale": [10, 10, 10]
+    },
+    {
+      "mesh": 1,
+      "name": "Sphere",
+      "scale": [0.5, 0.5, 0.5],
+      "translation": [5.605831146240234, 1.0160478353500366, 0],
+      "extensions": {
+        "KHR_audio": {
+            "emitter": 1
```
aaronfranke, line 85: This node has both an audio emitter and a Mesh. This significantly complicates import into Godot Engine which only allows one type per node, since we would need to import 2 nodes in this case.

Instead, I would be in favor of modifying the KHR_audio standard to only allow audio emitters on nodes that aren't meshes. Even aside from implementation difficulty, it makes more sense to have these be separate, so that an object's audio transform is not necessarily locked to the mesh transform.

DrX3D:
There is no ratified Audio extension. This PR will be closed without merge by 29 Nov. Future PRs should only use ratified extensions and be against Sample-Assets repo.

PR #361, opened 2022-10-05
Add MultipleScenes sample

javagl:
There currently is no sample model with multiple `scenes`.

I had created this example ~3 years ago (in the context of https://github.com/KhronosGroup/glTF-Sample-Models/issues/166#issuecomment-564770923 ), but it never made it into a PR due to the pending discussion to [Consider removing support for multiple scenes in a single asset](https://github.com/KhronosGroup/glTF/issues/1542) back then. 

However, multiple scenes _are_ currently allowed, and this is a **minimal** example for testing this. Direct link to the relevant directory: https://github.com/javagl/glTF-Sample-Models/tree/MultipleScenes/2.0/MultipleScenes


cx20:
I think the current `screenshot.png` probably uses the same one as the `Cameras` model screenshot.
https://github.com/javagl/glTF-Sample-Models/blob/MultipleScenes/2.0/MultipleScenes/screenshot/screenshot.png
https://github.com/javagl/glTF-Sample-Models/blob/MultipleScenes/2.0/Cameras/screenshot/screenshot.png

image

Below is the display result of Babylon.js. Perhaps it would be better to make a screenshot showing the square from the front like this.

https://sandbox.babylonjs.com/?assetUrl=https://rawcdn.githack.com/javagl/glTF-Sample-Models/dd9082554a09b20fd80e681aa5dd1567fcec4b3f/2.0/MultipleScenes/glTF/MultipleScenes.gltf

image

javagl:
@cx20 The screenshot was taken with my (old, Java-based) glTF viewer, and it might be the same as for the `Cameras` because it was just taken with "my (arbitrary) default camera". 

But I agree that it should be updated, and just did that with the screenshot that you provided.

(Maybe some screenshots from my `Simple*` models, and the `Cameras` one, should also be updated - but I'd rather do that in a dedicated PR)

---

I'm curious: I tried this sample with the viewer from DonMcCurdy and with Babylon, but did not _extensively_ try out _all_ viewers. You probably did that (or have a better idea there): Did you find **any** viewer that allows _selecting_ the scene that should be displayed? 

(The whole point of this sample is to make that clear: There _may_ be multiple scenes. I think that until now, viewers generally don't anticipate that...)

cx20:
@javagl Thanks for the updated screenshots.

> Did you find any viewer that allows selecting the scene that should be displayed?

I haven't come across a viewer that can change scenes, although I haven't tried that much yet.

BTW, Babylon.js seems to have a multi-scene concept, so it might be useful to be able to adapt it.

https://doc.babylonjs.com/features/featuresDeepDive/scene/multiScenes

javagl:
There's probably not sooo much controversial about this asset: It is supposed to be "the" minimal asset with multiple scenes. And currently, it's the _only_ asset that could allow implementors to check the behavior of their viewer for something that is explicitly allowed via the spec. 

We don't have a strict "review required" policy here, so unless one of the regular maintainers/contributors/mergers (@emackey, @lexaknyazev ...) disagrees, I'd just hit 'merge' here soon...

emackey (approved):

Issue #364, opened 2022-11-04
Normal texture TestMap_blur.png does not conform to the glTF 2.0 spec


erich666:
I thought this was an issue with the glTF 2.0 spec itself, but the problem is simply that the normal texture [TestMap_blur.png](https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/sourceModels/TextureTransformMultiTest/TestMap_blur.png) is poorly formed. There are a few of these PNGs used in this repo. They should be replaced with a proper texture, to avoid confusion (I certainly was confused to see it).

See https://github.com/KhronosGroup/glTF/issues/2223 for the full explanation and a possible replacement texture (one that is properly scaled in the blue, Z, channel).

Short version is that it maps blue [0.0 .. 1.0], not blue (0.5 .. 1.0], to z [0 .. 1]. For example, pixel 394,303 has a triplet of (242,89,82) - the blue value is too small, because they're using the wrong mapping. [The spec](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#additional-textures) is specific on this point:

> Normal textures **SHOULD NOT** contain blue values less than or equal to 0.5.

The triplet is valid if mapping the whole range of blue values to z [0 .. 1], but that's not what the spec requires.

Here's a full analysis of this triplet. It's easy to convert with both mappings and see if the value (242,89,82) uses one of them.

Assuming blue (0.5 .. 1.0]:
Vector 242 89 82 converts to 0.898, -0.302, -0.357
  length was wrong at 1.012

Assume blue [0.0 .. 1.0] (which is the wrong mapping, not what glTF requires):
Vector 242 89 82 converts to 0.898, -0.302, 0.322, length 1.001

Clearly this triplet is using the wrong mapping. When converted, it gives a vector of length 1.001 using this mapping, showing that this mapping is the intent of the texture.

erich666:
I made a replacement TestMap_Normal.png file, attached below.

Copied from https://github.com/KhronosGroup/glTF/issues/2223: Using [NormalMap-Online](https://cpetry.github.io/NormalMap-Online/), I negated [the TestMap_blur.png image](https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/sourceModels/TextureTransformMultiTest/TestMap_blur.png), negated the red channel (always needed), and set the z-range from -1 to +1 in the tool, then output. Here it is.
[...]

PR #365, opened 2022-11-07
Correct normal texture that has incorrect Z range

erich666:
See https://github.com/KhronosGroup/glTF-Sample-Models/issues/364


cx20:
I think it's probably better to correct the `glTF-Binary` file as well.

donmccurdy:
Updated `glTF-Binary` attached:

[TextureTransformMultiTest.glb.zip](https://github.com/KhronosGroup/glTF-Sample-Models/files/9992817/TextureTransformMultiTest.glb.zip)

emackey (approved):
Thanks @erich666.  I'll update the GLB on a separate branch.

PR #378, opened 2023-02-27
Adding sample for MSFT_texture_dds

Keenuts:
This sample was copied from Cube/, and used the glTF-Toolkit to export it with DDS textures.


emackey:
I think Khronos is trying to encourage the community to embrace the platform-independent KTX2 format with BasisU GPU compression instead of old DDS textures.  Most modern glTF readers don't support DDS.  It filled a gap in 2017 before KTX2+BasisU were available, but it's being phased out now.

Keenuts:
Hi! Thanks for the info! I agree KTX2 would be better in the long run as it's a more documented format, and is well integrated to gltf.
But DDS still has some benefits:
 - format is very simple -> upload to the GPU as-is (minus headers).
 - it's well supported by tools (gimp, imagemagick)
 - it's well supported on desktop
 vs KTX + basis which seem not to support storing BC directly, meaning there is a transcoding step.
 
 Digressing into DDS vs KTX2 :blush: 
 I can understand if you want to only keep samples which reflect what Khronos wants to push. Up to you!

bghgary:
I think this is the first PR that adds a model with vendor extensions. Do we want models with vendor extensions in this repo?

donmccurdy:
I think I would prefer a blanket policy — we probably do not want to host vendor extensions in this sample repository.

If there are multiple vendors implementing MSFT_texture_dds, then arguably it could be promoted to an EXT_ extension if a group of vendors are interested. I don't necessarily think Khronos should _lead_ that, for the reasons given above, but I also wouldn't block its existence in favor of KTX2. 

***

Aside — KTX2 can indeed store BC formats...

https://github.com/donmccurdy/KTX-Parse/blob/2b57a380fdc4525ca9fa1cb12e204f935ae2701c/src/constants.ts#L179-L194

... which can similarly be uploaded to the GPU directly, minus headers. I'd seen a tool for creating those KTX2 files somewhere, but can't find it at the moment. You're correct that it's not widely supported by tools though, and furthermore it wouldn't be allowed under the existing `KHR_texture_basisu` extension, which allows only Basis Universal formats.

Keenuts:
Seems that this PR won't make it as yes, it would bring vendor extensions in this repo, which seems not desirable.
Closing this, thanks for the feedback!

PR #380, opened 2023-03-02
Add test model for negative scale

emackey:
New test model.

~~Done: 2.0/README.md should be updated before merging this.~~


bghgary:
Would it be useful to link to the asset generator tests if people want to debug individual nodes?

https://github.com/KhronosGroup/glTF-Asset-Generator/tree/master/Output/Positive/Node_NegativeScale

